PR #3, opened 2022-05-09
Proposal 2: Project roadmap, May 2022

MarkLodato:
This is a proposed roadmap as presented on 2022-04-28. It is a draft and is open for discussion. It captures the work that I know about, but I'm sure there is more to be added.

### DIFF
--- a/0002/README.md
+++ b/0002/README.md
@@ -10,6 +10,9 @@
 This document proposes long-term themes and short-term milestones for the SLSA
 project.
 
+> NOTE TO READER: If you know of committed projects that should be added to this
+> roadmap, please send a pull request or file an issue in this repo.
+
 ## Implementation
 
 Once this proposal is accepted, we will track progress through GitHub Projects
@@ -61,7 +64,10 @@ foundation. Plans:
 -   Incorporate the concept of policy or verification into the specification,
     which is necessary to realize the security guarantees of the levels.
 
--   Address known clarity issues in v0.1, including updated terminology.
+-   Address known clarity issues from v0.1, including improved terminology
+    ([#306](https://github.com/slsa-framework/slsa/issues/306)), platform vs.
+    project requirements
+    ([#188](https://github.com/slsa-framework/slsa/issues/188)), and others.
 
 #### Milestone: Recommended SLSA provenance format has stable 1.0 release ([@MarkLodato], P2)
 
@@ -90,9 +96,9 @@ and that the SLSA properties are verified automatically before the software is
 delivered to the consumer.
 
 In the short term, we focus on open-source packaging ecosystems, tentatively
-starting with Python. Open-source is a more tractable problem than closed-source
-and serves as a good model by its open nature. We expect equivalent solutions
-for closed-source in the future.
+starting with Python.[^python-buy-in] Open-source is a more tractable problem
+than closed-source and serves as a good model by its open nature. We expect
+equivalent solutions for closed-source in the future.
 
 #### Milestone: SLSA 2-3 builds are possible for arbitrary open-source packages using several common build services
 
@@ -153,7 +159,10 @@ the producer. This work also includes aligning with SBOM, since the requirements
 are very similar. With this milestone, motivated consumers can act upon the
 provenance themselves.
 
-Design is P1 for [@lumjjb] in Q2.
+*   Q2: Design complete and approved within SLSA community ([@lumjjb], P1)
+*   Q3: Prototype implementation without requiring changes to PyPI
+*   Q4+: Python Enhancement Proposal (PEP) process, followed by implementation
+    if needed
 
 Issue: [#269](https://github.com/slsa-framework/slsa/issues/269)
 
@@ -183,12 +192,14 @@ Sub-problems:
     migrations.
 
 -   Implementation for each specific ecosystem, starting with Python/PyPI. This
-    includes a design, prototype, dashboard, and/or upload-time integration.
+    includes a design, prototype, dashboard, Python community outreach, PEP
+    authoring and approval, and/or upload-time integration.
 
--   (optional) Toy implementation of a mock package registry ("SLSA Playground")
-    to develop ideas and to show how SLSA would work in practice, without having
-    to change any existing system or to release any real package. This can also
-    be used as examples for https://slsa.dev/threats. (P3)
+-   (optional) Toy implementation of a
+    [mock package registry](https://github.com/jku/repository-playground) to
+    show how SLSA works in practice, without having to change any existing
+    system or to release any real package. This can also be used as examples for
+    https://slsa.dev/threats. (P3)
 
 #### Milestone: Client-side verification of SLSA policies in the Python ecosystem removes the need to trust the package registry (P3)
 
@@ -206,8 +217,13 @@ Widescale adoption of SLSA across open-source is the long-term goal, but it is
 not practical in 2022 because it not only depends on the work above, but also
 work to minimize or eliminate any friction for project maintainers.
 
+#### Milestone: Early adoption by key projects raises awareness of SLSA (P2)
+
 In the short term, we hope to have targeted adoption of key open-source projects
-to both exemplify standards and to reduce risk for those projects.
+to both exemplify standards and to reduce risk for those projects. We can also
+publish adoption stories to inform readers where SLSA is being applied. Finally,
+we can reach out to open-source communities, working groups, conferences, etc.
+to raise awareness of SLSA.
 
 ### Theme: Government and industry
 
@@ -221,7 +237,16 @@ standards, such as SSDF, SBOM, etc. Exactly which government standards is TBD.
 -   Alignment between SLSA and SSDF is clearly documented through a compelling
     story of synergy. ([@hepwori], P2)
 
-<!-- Footnotes and links -->
+## Footnotes
+
+[^python-buy-in]: We plan to reach out to the Python community to get buy-in and
+    go through the official Python Enhancement Proposal (PEP)
+    process. At the time of writing (May 2022), this has not yet
+    started so any reference to Python or PyPI in this document
+    should be considered a placeholder in case we choose a
+    different ecosystem down the road.
+
+<!-- Link definitions -->
 
 [@MarkLodato]: https://github.com/MarkLodato
 [@bobcallaway]: https://github.com/bobcallaway
--- a/0002/README.md
+++ b/0002/README.md
@@ -1,6 +1,6 @@
 # Proposal 2: Project roadmap, May 2022
 
-*   Proposer: [Mark Lodato](https://github.com/MarkLodato) (lodato@google.com)
+*   Proposer: [Mark Lodato][@MarkLodato] (lodato@google.com)
 *   GitHub Issue: n/a
 *   Status: [DRAFT](../README.md#meaning-of-status-codes)
 *   Implementation: TODO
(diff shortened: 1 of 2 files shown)


0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+-   Pare SLSA v1.0 down to only cover build integrity, which is SLSA's core
+    strength. This makes SLSA practical by removing the requirements that cannot
+    be implemented today, notably the controversial two-party review requirement
+    and the undefined "common" requirements.
+
+-   Establish a convention for extending SLSA beyond build integrity, such as to
+    vulnerability management or source integrity. This might be through multiple
+    independent ladders, or a single ladder, or something else entirely. Note:
+    this is just about establishing the convention; actually extending SLSA is
+    out of scope for v1.0.
```
joshuagl: We should not rule out a (small) extension within the v1.0 timeframe. I think it may make sense to validate the extension convention.

0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+[provenance]: https://slsa.dev/provenance
+[slsa spec]: https://slsa.dev/spec
+
+#### Milestone: SLSA community is healthy according to some agreed-upon definition
+
+The SLSA project has an agreed-upon definition of community health, with regular
+measurements and a plan to improve it over time. This likely includes the number
+of contributors and the lack of barriers for new contributions. Ideally there
+should be a clear path for potential contributors to jump in and start helping
+with SLSA.
```
joshuagl, line 88: 💯 to all of this.

0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+    healthy surrounding community.
+-   **Tooling**, services, and documentation make SLSA readily adoptable.
+-   **Adoption** of SLSA by open-source projects reduces security risk for
+    users.
+-   **Government and industry** widely accept SLSA as the *lingua franca* of
+    supply chain security.
+
+## Milestones (short-term)
+
+The following milestones are expected to be practical in the next few quarters.
```
joshuagl, line 36: There's quite a chunk of work exciting work here – do we have some known willing contributors somewhere? I can't imagine the current active contributors to SLSA would be able to achieve all of this, even if it were their full-time job. 

Some of these milestones have dependencies on other communities – have we engaged with them and sought their buy-in before making their adoption a milestone?

Should we hold ourselves accountable to these milestones. How will we do that?

0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+-   Establish a convention for extending SLSA beyond build integrity, such as to
+    vulnerability management or source integrity. This might be through multiple
+    independent ladders, or a single ladder, or something else entirely. Note:
+    this is just about establishing the convention; actually extending SLSA is
+    out of scope for v1.0.
+
+-   Incorporate the concept of policy or verification into the specification,
+    which is necessary to realize the security guarantees of the levels.
+
+-   Address known clarity issues in v0.1, including updated terminology.
```
joshuagl: Specifically, and implicit in some of the milestones below, we should clarify things enough for platforms to implement what they need to (platform vs. project requirements https://github.com/slsa-framework/slsa/issues/188).

0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+-   Policy model, which securely maps package name to canonical source
+    repository. For example, [requests](https://pypi.org/project/requests/) maps
+    to https://github.com/psf/requests. This metadata already exists for some
+    packages, but (a) it is not structured and (b) package maintainers can
+    mutate it at will. The solution must protect against malicious changes while
+    silently allowing official changes in source repository. One idea is to have
+    a Trust On First Use (TOFU) model with a mechanism for source repo
+    migrations.
+
+-   Implementation for each specific ecosystem, starting with Python/PyPI. This
```
joshuagl, line 196: we mention Python/PyPI explicitly here but talk about PyPI/Python or npm above -- is Python the chosen representative implementation? Should we remove references to npm?

0002/README.md
```diff
@@ -0,0 +1,209 @@
...
+This probably won't happen in 2022, but we list it here for completeness.
+
+### Theme: Adoption
+
+Widescale adoption of SLSA across open-source is the long-term goal, but it is
+not practical in 2022 because it not only depends on the work above, but also
+work to minimize or eliminate any friction for project maintainers.
+
+In the short term, we hope to have targeted adoption of key open-source projects
+to both exemplify standards and to reduce risk for those projects.
```
joshuagl: An adoption milestone for this roadmap might be to share about SLSA and the work we're doing at appropriate forums for the representative implementation (i.e. WG's, conferences, etc).

We could also start capturing adoption stories (for many projects a "friends" repository has become a standard way to document adoption) – for example the recent Tekton case study.

joshuagl (approved):
Thanks Mark. This is a very solid proposal for paving the road to SLSA adoption. I'm excited to work with folks on this!

I've left a few comments inline, but broadly this looks good to merge as a draft. My only major concern is that this feels like a lot of work and I have no idea who will do it all 😄

0002/README.md
MarkLodato: Reworded. How does that look?

0002/README.md
MarkLodato: Removed NPM. (We might also propose something for NPM if time permits, but no need to list that here.)

We are working with @di, who is a PSF board member. He's on board with the idea in general, without committing anything. That said, we still have to discussed it with the Python community and go through the PEP process. My thinking was that we want to start with a design and prototype implementation to give us something concrete to discuss.

0002/README.md
MarkLodato: Added owners and milestones. PTAL.

0002/README.md
joshuagl: Looks great, thanks!

0002/README.md
joshuagl: Rephrasing looks good, thanks. I think we (SLSA) should be open about what level of buy-in we have (lest folks assume this is a done-deal). But that might fit as part of the detailed planning i.e., listing Python outreach and PEP authoring as tasks?

0002/README.md
```diff
@@ -0,0 +1,232 @@
...
+    packages, but (a) it is not structured and (b) package maintainers can
+    mutate it at will. The solution must protect against malicious changes while
+    silently allowing official changes in source repository. One idea is to have
+    a Trust On First Use (TOFU) model with a mechanism for source repo
+    migrations.
+
+-   Implementation for each specific ecosystem, starting with Python/PyPI. This
+    includes a design, prototype, dashboard, and/or upload-time integration.
+
+-   (optional) Toy implementation of a mock package registry ("SLSA Playground")
```
joshuagl: A colleague at VMware (👋 @jku) is starting a mock package registry project – https://github.com/jku/repository-playground – with a goal of providing a place to develop ideas related to the OpenSSF [Securing Software Repos WG](https://github.com/ossf/wg-securing-software-repos).

It would be fantastic if we could share mock package registry implementation to enable folks to focus more on the security problems they are trying to solve.

0002/README.md
joshuagl: This is amazing commitment to the project from Google, thank you!

I am looking forward to seeing more companies represented as owners as the project continues to grow (and owning some milestones myself).

joshuagl (approved):

0002/README.md
MarkLodato: Awesome! Great minds think alike! 😉  I'd love to use this to show how SLSA works! Added a link to the doc.

0002/README.md
MarkLodato: Added, though I'm not 100% confident in what I wrote. PTAL.

0002/README.md
MarkLodato: Good point. Added.

0002/README.md
MarkLodato: Added.

0002/README.md
MarkLodato: Sounds good. Added a note at the top. Want to send a follow-up PR adding stuff from your side? Also I tried to word it to make it clear that any persons identified are "leads" but not the sole party doing the work. For all of these designs and implementations, we plan to work in an open fashion.

0002/README.md
```diff
@@ -0,0 +1,257 @@
...
+approach is that it allows us to add SLSA without requiring upstream changes to
+the build service, at the cost of increasing the size of the trusted computing
+base and possibly worse user experience than if the builder had provided SLSA
+directly. This allows motivated software producers to achieve SLSA immediately.
+Planned Sigstore-based solutions include:
+
+-   [GitHub Actions](https://github.com/features/actions) -
+    [slsa-github-generator](https://github.com/slsa-framework/slsa-github-generator)
+    ([@laurentsimon], P0)
+-   [GitLab CI/CD](https://about.gitlab.com/topics/ci-cd/) ([@bobcallaway], P2)
```
abacchi: Relevant Gitlab issue: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/28940

0002/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+Widescale adoption of SLSA across open-source is the long-term goal, but it is
+not practical in 2022 because it not only depends on the work above, but also
+work to minimize or eliminate any friction for project maintainers.
+
+#### Milestone: Early adoption by key projects raises awareness of SLSA (P2)
+
+In the short term, we hope to have targeted adoption of key open-source projects
+to both exemplify standards and to reduce risk for those projects. We can also
+publish adoption stories to inform readers where SLSA is being applied. Finally,
```
abacchi, line 226: I really like this milestone- I see these adoption stories as great way to promote SLSA, show some of the techniques used to get there, and increase the security of the projects.

0002/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+#### Milestone: SLSA specification has a stable 1.0 release ([@MarkLodato], P1)
+
+Version 1.0 of the [SLSA specification](https://slsa.dev/spec) provides a stable
+foundation. Plans:
+
+-   Pare SLSA v1.0 down to only cover build integrity, which is SLSA's core
+    strength. This makes SLSA practical by removing the requirements that cannot
+    be implemented today, notably the controversial two-party review requirement
+    and the undefined "common" requirements.
```
abacchi, line 56: Is this proposing to remove the two-party review from SLSA level 4 + others?

0002/README.md
MarkLodato: Thanks! Added. Also added @samwhite-gl's name.

0002/README.md
MarkLodato: Yes, that is what I plan to propose, which will need to be approved independently of this roadmap. We'll have to see if the new proposed ladder makes sense overall. Two-party review would instead become an "extension" under the next bullet, possibly as part of a larger effort to ensure the code is "good". This might also mean adding other alternatives to achieve the same goal.

Today, mandatory two-party review is impractical for most organizations, particularly open-source projects:
- Most organizations lack the culture, tooling, and infrastructure to maximize review effectiveness, minimize overhead, and ensure strong security guarantees.
- We don't yet have technical means to verify two-party review happened on any popular code hosting sites, such as GitHub or GitLab *(edit: see following comments for clarification)*, and we don't want to block the build integrity piece of SLSA on this.

I plan to write this all up as part of the proposal. 

Are you suggesting any changes to the text, or just asking for background?

0002/README.md
joshuagl: LGTM, thank you.

0002/README.md
joshuagl: LGTM, thanks!

0002/README.md
samwhite-gl: > We don't yet have technical means to verify two-party review happened on any popular code hosting sites, such as GitHub or GitLab, and we don't want to block the build integrity piece of SLSA on this.

@MarkLodato I don't know about GitHub, but FWIW GitLab actually has several ways of verifying that two-party review happened.  First, MR approval rules can be configured to [require two-party review](https://docs.gitlab.com/ee/user/project/merge_requests/approvals/rules.html) with an option to [disallow authors and committers from approving](https://docs.gitlab.com/ee/user/project/merge_requests/approvals/settings.html).  Second, whenever review happens, it is recorded in the Audit Log and can be verified in that way.  For a more convenient and visual verification method, a [Compliance Report](https://docs.gitlab.com/ee/user/compliance/compliance_report/#separation-of-duties) is available that flags any MRs that did not properly go through two-person review or if the author or a committer to the MR self-approved their own code.

MarkLodato: Thank you for the correction, @samwhite-gl! I wasn't aware of this feature. Let me try to rephrase that bullet:

* *Verifying that two-party review happened will require more work to make it practical at scale*, and we don't want to block the build integrity piece of SLSA on this.

In particular::

* Most projects have some amount of historical unreviewed commits. How should SLSA handle this? We could require all these old commits to get reviewed, but that is impractical and of dubious value. We could require review of all old commits within the past X days to get reviewed, but that has the same problem on a smaller scale. We could require all new commits going forward to get reviewed, but then what if they turn it off again, or what if they turn it on right before building? I don't see an obvious best answer here.
* To scalably use the GitLab information, I think we'd need it in the form of a signed attestation rather than making the verifier call the API every time. So that piece would need to be built. That's also true of build provenance, but decoupling the two means that we can get build integrity done before tackling two-party review.

abacchi: @MarkLodato It was more just to spur a conversation like the one above! :) thanks. I agree with you on the above points as well.

MarkLodato:
Thanks all! I'll submit now but as par our process, this is still a DRAFT and open to more changes. I'll also file a GitHub Issue in the main SLSA repo to track acceptance.

PR #5, opened 2022-06-15
Update proposal 0002 (roadmap) and mark as accepted.

### LINKED ISSUE #384: Project Roadmap, May 2022
This issue tracks acceptance of [Proposal 2: Project roadmap, May 2022](https://github.com/slsa-framework/slsa-proposals/blob/main/0002/README.md).

If you have any questions or comments on the plan, feel free to add them here and I can respond or update the roadmap. Pull requests also welcome.

I suggest that we accept the roadmap in about a week if there are no significant concerns.

MarkLodato:
- Update dates to end of year, given that we're already past mid-2022.
- Avoid over-specifying the v1.0 plans since some things are not yet agreed upon.
- Clarify the "policy" bullet to avoid a common source of miscommunication.
- Remove two bullets (community health and mock package registry) that have been deprioritized.
- Mark the proposal as accepted, given the lack of disagreement on the issue.

Fixes https://github.com/slsa-framework/slsa/issues/384.

### DIFF
--- a/0002/README.md
+++ b/0002/README.md
@@ -39,7 +39,7 @@ Each item or sub-item has a priority and a main point of contact.
 Priorities are to be interpreted as follows:
 
 -   P0: High confidence it will happen mid-2022
--   P1: Expected mid-2022 but may slip
+-   P1: Expected by end of 2022
 -   P2: Planned for 2022 but may slip or get de-prioritized
 -   P3: Side project, moderate risk of de-prioritization
 
@@ -50,10 +50,11 @@ Priorities are to be interpreted as follows:
 Version 1.0 of the [SLSA specification](https://slsa.dev/spec) provides a stable
 foundation. Plans:
 
--   Pare SLSA v1.0 down to only cover build integrity, which is SLSA's core
-    strength. This makes SLSA practical by removing the requirements that cannot
-    be implemented today, notably the controversial two-party review requirement
-    and the undefined "common" requirements.
+-   Pare SLSA v1.0 down to only the levels and requirements that are generally
+    agreed upon and practical to implement, leaving others for future extension.
+    In particular, v1.0 will tentatively only define SLSA 1-3 at first, marking
+    SLSA 4 as "draft" while the community reworks the two-party review and
+    hermeticity requirements.
 
 -   Establish a convention for extending SLSA beyond build integrity, such as to
     vulnerability management or source integrity. This might be through multiple
@@ -61,8 +62,10 @@ foundation. Plans:
     we will validate the framework through at least one draft extension, but
     finalizing the first extension is not required for v1.0.
 
--   Incorporate the concept of policy or verification into the specification,
-    which is necessary to realize the security guarantees of the levels.
+-   Incorporate the concept of "policy" into the specification, whereby
+    downstream systems only accept artifacts whose provenance matches some
+    expected patterns. This is necessary to realize the security guarantees of
+    the levels.
 
 -   Address known clarity issues from v0.1, including improved terminology
     ([#306](https://github.com/slsa-framework/slsa/issues/306)), platform vs.
@@ -78,14 +81,6 @@ Requirements for v1.0 are still TBD.
 [attestation]: https://github.com/in-toto/attestation
 [dsse]: https://github.com/secure-systems-lab/dsse
 [provenance]: https://slsa.dev/provenance
-
-#### Milestone: SLSA community is healthy according to some agreed-upon definition ([@hepwori], P2)
-
-The SLSA project has an agreed-upon definition of community health, with regular
-measurements and a plan to improve it over time. This likely includes the number
-of contributors and the lack of barriers for new contributions. Ideally there
-should be a clear path for potential contributors to jump in and start helping
-with SLSA.
 
 ### Theme: Tooling
 
@@ -196,12 +191,6 @@ Sub-problems:
 -   Implementation for each specific ecosystem, starting with Python/PyPI. This
     includes a design, prototype, dashboard, Python community outreach, PEP
     authoring and approval, and/or upload-time integration.
-
--   (optional) Toy implementation of a
-    [mock package registry](https://github.com/jku/repository-playground) to
-    show how SLSA works in practice, without having to change any existing
-    system or to release any real package. This can also be used as examples for
-    https://slsa.dev/threats. (P3)
 
 #### Milestone: Client-side verification of SLSA policies in the Python ecosystem removes the need to trust the package registry (P3)
 
--- a/0002/README.md
+++ b/0002/README.md
@@ -2,7 +2,7 @@
 
 *   Proposer: [Mark Lodato][@MarkLodato] (lodato@google.com)
 *   GitHub Issue: [#384](https://github.com/slsa-framework/slsa/issues/384)
-*   Status: [DRAFT](../README.md#meaning-of-status-codes)
+*   Status: [ACCEPTED](../README.md#meaning-of-status-codes)
 *   Implementation: TODO
 
 ## Abstract
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ should instead live in the [main SLSA repo].
 Number                 | Title                        | [Status]
 ---------------------- | ---------------------------- | -----------
 [0001](0001/README.md) | Creation of proposal process | IMPLEMENTED
-[0002](0002/README.md) | Project roadmap, May 2022    | DRAFT
+[0002](0002/README.md) | Project roadmap, May 2022    | ACCEPTED
 
 ## How to create a proposal document
 


JanZerebecki:
Please do https://github.com/slsa-framework/slsa-proposals/pull/6 before this.

MarkLodato:
I updated this PR to clarify some bullets, hopefully avoiding the confusion from #6.

joshuagl (approved):

JanZerebecki (changes requested):
Mark, you didn't answer the questions from what I said in https://github.com/slsa-framework/slsa-proposals/pull/6 and https://github.com/slsa-framework/slsa/issues/384 . Instead just closed the PR. Let me prefix the underlying general question:

Is end-to-end supply chain integrity in scope for SLSA?

I think SLSA should answer this before finalizing plans for 1.0. SLSA in its current version to my understanding implicitly rejects end-to-end integrity (which would allow offline verification and stored integrity evidence for forensic analysis and resistance to full compromise even in the face of the compromise of the machines of multiple involved engineers) in favor of transient hop-to-hop integrity (which does not store evidence and thus doesn't allow offline verification nor forensic analysis and has central single points of failure).

Thus the more specific questions:
Will SLSA accept changes to suggest in each requirement end-to-end techniques (while still allowing hop-to-hop ones if those are easier to adopt)?
Will SLSA accept to include in draft level 4 or higher equivalent requirements that only allows end-to-end techniques replacing the hop-to-hop ones in lower levels?

JanZerebecki:
Maybe @slsa-framework/slsa-steering-committee as a whole also wants to answer my above 3 questions, as these are important to the direction of SLSA.

MarkLodato:
@JanZerebecki, please excuse my directness, but I do not understand at all what you mean by end-to-end integrity vs hop-to-hop integrity. What specific security threats is SLSA missing now, and how would your suggestions address those? How would it actually work in practice? Can you work through a real-world example to illustrate? Is it practical for a significant number of organizations to adopt? As a next step, I suggest writing a more detailed proposal that answers these questions. It doesn't have to be long, and starting with an example might be the quickest path forward.

I am sure that the SLSA community is open to suggestions for how to improve, but we need more detail, and major changes to SLSA 1-3 are unlikely to happen in v1.0. The tentative plan is to only define SLSA 1-3 in v1.0 and defer definition of SLSA 4 until v1.1. (SLSA 4 has some issues we want to address, but we don't want to block v1.0 for that.)

JanZerebecki:
@MarkLodato Please be direct and frank with me if it helps improve open source security.

Hop-to-hop integrity: Alice pushes to repo foo/bar.git on Github with an ssh key. Github check the ssh-key authorization with their non-public permissions of the repo. Bob pulls from foo/bar.git while checking the TLS cert from the HTTPS connection.

End-to-end integrity examples (see https://github.com/slsa-framework/slsa/issues/421 for the issue asking to require something like these in level 4+):
1) Alice signs her commit with GPG. Pushes to PR. Carol pushes signed merge of PR to master branch. Bob pulls and checks that the top commit is signed by a key in a list for this repo.
2) Alice releases Rust package bar and then uses cargo-crev to review and sign it. Carol and Dave also review and sign with cargo-crev. Bob checks that there are 3 trusted reviews with cargo-crev and updates to the new version of bar.

These work today and are in use by some open source projects. Yes, these tools can be improved, but are still better than only hop-to-hop. Yes, it is possible for a significant number of organizations to adopt these. I can give examples of ones that did.

Concentrationg here on SLSA 0.1 threat A, B, F, G. Example threats (I have more):
1) Central authorization provider, registry or code host is compromised.
2) Domain registration of one of the above or of the mail provider of a project maintainer is transferred to malicious actor. The actor obtains a new TLS cert, which then is valid. (Something similar happened with npm package foreach: https://nitter.net/lrvick/status/1523787247706951680 .)
3) AFAIK Github still has a process in place to deactivate second factor on an account, by something that is usually where the first factor also is. This could be used to take over the account after the first factor was compromised. (SLSA threat A2 point 4)
4) Someone uses SMS as a second factor. There are many examples of compromises where this was defeated. (SLSA threat A2 point 4)

I have published a more high level description of threat actor capabilities I'm working from, if you are interested. In addition https://arxiv.org/abs/2204.04008 is very useful. I would like to extend the thread model explained in SLSA based on these and fill in some of the TBD in there, if that will be allowed. I have not proposed any PR directly yet as the threat model document currently reads as these being out of scope.

Is this sufficient for you to understand? Does that answer what you asked for?

> Will SLSA accept changes to suggest in each requirement end-to-end techniques (while still allowing hop-to-hop ones if those are easier to adopt)?

> > and major changes to SLSA 1-3 are unlikely to happen in v1.0

Do you mean that as "no, not before 1.0" to my above question?

MarkLodato:
Thanks, yes, that is very helpful!

First, the roadmap is not a good place to discuss this. Instead, I suggest that we move this to the [Specification SIG](https://docs.google.com/document/d/1kMP62o3KI0IqjPRSNtUqADodBqpEL_wlL1PEOsl6u20/edit). It would be great if you could attend the meeting, which is Mondays at 12-1pm US Eastern. Otherwise, perhaps via Slack (openssf.slack.com #slsa-specification) or GitHub Issue.

But to quickly answer your specific questions:

> Alice signs her commit with GPG. Pushes to PR. Carol pushes signed merge of PR to master branch. Bob pulls and checks that the top commit is signed by a key in a list for this repo.

IIUC this is about source integrity, i.e. protection against tampering of the sources. (nit: I don't think the term "end-to-end" is accurate because the other "end" is the source code, not the final binary. In other words, there are still two hops: commit signing from developer to source, and then provenance from source to binary.) With respect to SLSA, commit signing could be one way to implement the [Verified identity](https://slsa.dev/spec/v0.1/requirements#verified-history), and I agree that it has the desirable property of removing the source control system from the trust base. It seems worthwhile to consider whether this should be a requirement at some level.

But for v1.0, we are _considering_ moving in the opposite direction by splitting out source requirements entirely to their own level. In other words, instead of adding more source requirements, we would remove them entirely and then add a new thing that describes desirable properties of the source. That would be a good venue for this discussion. (To be clear, this is just an idea at this point - no decisions have been made.)

> Alice releases Rust package bar and then uses cargo-crev to review and sign it. Carol and Dave also review and sign with cargo-crev. Bob checks that there are 3 trusted reviews with cargo-crev and updates to the new version of bar.

It's less clear to me how this would fit cleanly into the SLSA framework. I agree that it could be a good thing. Have to think more about it.

> > Will SLSA accept changes to suggest in each requirement end-to-end techniques (while still allowing hop-to-hop ones if those are easier to adopt)?
> 
> > > and major changes to SLSA 1-3 are unlikely to happen in v1.0
> 
> Do you mean that as "no, not before 1.0" to my above question?

Correct, that is my take, but feel free to suggest this to the group.

JanZerebecki:
Thank you.

Sorry to have this discussion here, but I tried and failed multiple times elsewhere. And here seems to work finally. I'll focus only on the roadmap.

If the source requirements are not included in 1.0 then I'm fine with that as we can discuss what I think needs to be tweaked there after 1.0.

I have one or two more points about the roadmap outside of the requirements, but I need a few more days to properly ask about those. I hope that is ok.


There is then only one remaining point in level 1-3 of the requirements that I don't think is OK for 1.0: The Non-falsifiable provenance requirement is as written too early in the levels compared to how 1.0 will look now. And the technical requirements are not properly explained or not detailed enough or misleading. Can we also remove that for 1.0 and work on it afterwards?

Details for why Non-falsifiable needs to be fixed:
Rerunning the providence generator is easier than the build itself. It should be possible to generate providence in a reproducible way. Both economically and security wise it doesn't make sense to sign anything for security purposes before you have put work into verifiable reproducible builds. If you don't have evidence for a good build and just sign something, one might think: oh there is a signature, must be secure. Even when you have reproducible builds, so that you have some falsifiable evidence for non-compromised builds, then: Signing keys in a "secure management system appropriate for such material" is under specified. Most of the software to sign stuff is architecturally less secure than GPG and TUF, so if we go by the average understanding, that will lead people to put work in adopting worse. Even TUF can be used, with some effort, in a way, that is not a good idea, so we need to say what we mean. Etc.

MarkLodato:
> I have one or two more points about the roadmap outside of the requirements, but I need a few more days to properly ask about those. I hope that is ok.

Sure, we can always do a follow-up PR.

> There is then only one remaining point in level 1-3 of the requirements that I don't think is OK for 1.0: The Non-falsifiable provenance requirement is as written too early in the levels compared to how 1.0 will look now. And the technical requirements are not properly explained or not detailed enough or misleading. Can we also remove that for 1.0 and work on it afterwards?

Please file an issue in the main slsa repo. We will tag it as v1.0 and discuss it along with all the potential other v1.0 changes.

PR #6, opened 2022-07-06
0002: for 1.0 require verification for higest level

JanZerebecki:
See https://github.com/slsa-framework/slsa/issues/384#issuecomment-1175769763

### DIFF
--- a/0002/README.md
+++ b/0002/README.md
@@ -64,6 +64,16 @@ foundation. Plans:
 -   Incorporate the concept of policy or verification into the specification,
     which is necessary to realize the security guarantees of the levels.
 
+    Policy here means something for which reviewable evidence that can be
+    automatically mirrored is provided for any human to review.
+    Verification, which we favor, means one which anyone can run a machine
+    that will do that by itself (e.g. we'll make verified reproducible builds
+    required in the highest level).
+    We will try to avoid it but some policy might be indirect, meaning a
+    requirement on the policy for contribution to the source.
+    Only in lower levels will these categories be relaxed by trading in
+    verifiability in favor of easy security gains.
+
 -   Address known clarity issues from v0.1, including improved terminology
     ([#306](https://github.com/slsa-framework/slsa/issues/306)), platform vs.
     project requirements


0002/README.md
```diff
@@ -63,6 +63,9 @@ foundation. Plans:
 
 -   Incorporate the concept of policy or verification into the specification,
     which is necessary to realize the security guarantees of the levels.
+    Ensure that it is sufficient and necessary that all requirements to
+    achieve the highest level (5?) are end-to-end unasisted machine verified
+    and can be automatically redone independently.
```
mlieberman85: Can you elaborate what this means? end-to-end unassisted machine verification of policy makes sense, but it's not clear what "automatically redone independently" means here. Does it just mean validation should be able to be performed by multiple systems?

JanZerebecki: Yes, so that there is no single point of failure. See e.g. https://github.com/in-toto/rebuilderd for a validator of reproducibility. A validator needs to be easy to setup and keep running. It follows the changes of the source and reproduces the verification and attests to that by signature. Any client, like a software updater, before making use of the binaries, waits for a threshold of chosen verifiers.

Does that explain it?

Do you think my wording should be improved? Do you have a suggestion?

MarkLodato: Sorry, I don't understand the intent is of this change. Could you explain? The purpose of this roadmap is to give the reader a rough idea of the broad strokes of the changes we're considering. It is not intended to be complete or nuanced. Do you feel that this change rises to that level?

I would lean towards discussing this in the detailed v1.0 proposal but omitting it here, unless it's important for readers to understand.

JanZerebecki: Yes, I feel it rises to that level.

There can be different ways to check compliance to a specification: none, classical human institutions, cryptographic protocol, etc. My suggested change clarifies in which direction to change SLSA. Without it the reader has no idea about the rough direction.

It will take many month of my work time, so it is a broad stroke.

MarkLodato: Any other opinions?

Personally I don't understand how these words would map to practical concepts in the specification, so it just adds confusion for me. Again, I'm happy to discuss in the specification, but I'm not sure this is adding clarity to the roadmap.

JanZerebecki: Above I mentioned https://github.com/slsa-framework/slsa/issues/5 with rebuilderd, I explained another one in: https://github.com/slsa-framework/slsa/issues/421

Are these sufficiently practical concepts in the spec? Did that make the proposed words understandable? Do you have an idea what words would be better to add clarity?

JanZerebecki: I noticed that I neglected to add the following:

There will remain some policy we want to retain that will require human review to verify. For these the evidence needs to be referenced in a way that a validator can automatically mirror it. A few rare cases might remain where the policy is only indirect as a requirement on the policy for contributing to the source like a DCO.

Maybe a better alternative to this PR would then be:

Policy here means something for which reviewable evidence that can be automatically mirrored is provided for any human to review. Verification, which we favor, means one which anyone can run a machine that will do that by itself. We will try to avoid it but some policy will be indirect, meaning a requirement on the policy for contribution to the source. Only in lower levels will these categories be relaxed by trading in verifiability in favor of easy security gains.

Is that better?

MarkLodato:
I am going to close this PR in favor of #5, which I updated to hopefully address this concern. In particular, (1) I explained what "policy" meant and (2) tried to go into a little less detail about what v1.0 will encompass.

PR #8, opened 2023-02-28
Policy Architecture for Npm Package Provenance

wietse-gmail:
@MarkLodato I'm still doing light edits for readability and some appendix cleanups.  Meanwhile I have a few questions:
*   What do we do with the box with reviewers and LGTMs? 
*   I left some TBDs in  metadata fields in REAMD.md and 0004/REAMDE.md What is appropriate?


0004/README.md
```diff
@@ -0,0 +1,522 @@
+# Policy Architecture for Npm Package Provenance
+
+*   Proposer: Wietse Venema ([wietse@gmail.com](mailto:wietse@gmail.com))
+*   GitHub Issue: TODO
+*   Status: TODO
```
MarkLodato:
```suggestion
*   Status: [DRAFT](../README.md#meaning-of-status-codes)
```

MarkLodato (approved):
> What do we do with the box with reviewers and LGTMs?

Any thoughts @slsa-framework/slsa-steering-committee? I'm OK leaving it but don't feel strongly.

/cc @kommendorkapten @feelepxyz @mlieberman85 (previous reviewers)

trishankatdatadog:
> Any thoughts @slsa-framework/slsa-steering-committee? I'm OK leaving it but don't feel strongly.

Sorry, are you asking us to review this PR, or is it a specific Q about the table of reviewers in the MD?

P.S. This proposal reminds me of something @fridex [wrote](https://discuss.python.org/t/pre-pep-recording-provenance-of-installed-packages/23340) for the provenance of packages installed with pip...

0004/README.md
```diff
@@ -0,0 +1,522 @@
...
+An enforcing policy evaluator would have to be integrated with "npm install" so
+that it can block the installation of a non-compliant package.
+
+## References
+
+#### Lock file-based package management systems
+
+-  [https://blog.shalvah.me/posts/understanding-lock files](https://blog.shalvah.me/posts/understanding-lockfiles)
+    covers lock files in php, jode.js (with npm and yarn), and in go and ruby.
+-  [https://pypi.org/project/pipfile/0.0.2/](https://pypi.org/project/pipfile/0.0.2/)
```
fridex, line 415: Not sure if relevant, but please note Python still does not have a standard for lockfiles (AFAIK Pipfile is not standardized in the Python packaging).

MarkLodato:
> Sorry, are you asking us to review this PR, or is it a specific Q about the table of reviewers in the MD?

I guess both :).

joshuagl:
> > What do we do with the box with reviewers and LGTMs?
> 
> Any thoughts @slsa-framework/slsa-steering-committee? I'm OK leaving it but don't feel strongly.

I think it's clearer to readers of the repository if we remove them (as they don't exist in other proposals). We should ask the same folks to review the proposal here, too.

0004/README.md
wietse-gmail: Thanks. I'll add a note to this reference, after I figure out this user interface.

wietse-gmail:
> > > What do we do with the box with reviewers and LGTMs?
> > 
> > 
> > Any thoughts @slsa-framework/slsa-steering-committee? I'm OK leaving it but don't feel strongly.
> 
> I think it's clearer to readers of the repository if we remove them (as they don't exist in other proposals). We should ask the same folks to review the proposal here, too.

Thanks. I'll delete the box and work with Mark on who to ask for review.

trishankatdatadog:
I'm happy to review 🙂

wietse-gmail:
Thanks. I'm doing a quick sanity check pass and will flag it as 'ready'.

wietse-gmail:
@trishankatdatadog the text is ready for review!

wietse-gmail:
I flagged this PR as ready for review.

MarkLodato:
@kommendorkapten @joshuagl would you mind approving this PR since you LGTM'd the original doc? Thanks!

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+the evaluator will need to be integrated into the package installer, and
+expectations may become managed centrally. 
+
+### Generating provenance expectations
+
+To overcome a bootstrapping problem, we assume that compromises will be rare,
+and propose to generate package-specific SLSA provenance expectations on first
+use. Each such ‘trust on first use' policy contains expectations based on
+observed provenance, and can be used as a baseline to detect unexpected changes,
+for example a too low SLSA level, missing provenance, or or an unexpected change
```
joshuagl:
```suggestion
for example a too low SLSA level, missing provenance, or an unexpected change
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+in a ‘sensitive' provenance property.
```
joshuagl:
```suggestion
in a 'sensitive' provenance property.
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+package user. We also expect that policies will initially be alerting only, i.e.
+not enforcing.
+
+As experience grows and confidence increases, we expect that policies will
+evolve to become ‘enforcing', and that expectations may become managed
+centrally.
+
+## Background: npm 
+
+[Npm](https://www.npmjs.com/) is the package manager for Node.js software
```
joshuagl:
```suggestion
[npm](https://www.npmjs.com/) is the package manager for Node.js software
```
AIUI npm is never capitalised.

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+packages. Of particular relevance for SLSA provenance policy evaluation are the
+_package provenance_, _package content_, and (after installing a package) the
+_package lock file_. We summarize main features and present some numbers based
+on an analysis of the top-1000 npm packages ranked by the number of
+dependents.
+
+#### Package provenance
+
+-  [Npm RFC 626](https://github.com/npm/rfcs/pull/626) proposes to generate
```
joshuagl:
```suggestion
-  [npm RFC 626](https://github.com/npm/rfcs/pull/626) proposes to generate
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+    versions, and digests.
+-  Less than 1% of the top-1000 npm packages contain a package-lock.json
+    file. If the file is present, it will by default not affect what dependency
+    versions "npm install" will install.
+
+## Implementing provenance expectations 
+
+There are multiple ways to implement expectations for provenance properties that
+are not expected to change (including properties that are inferred from
+provenance): do we remember ‘good' property values, or do we remember ‘good'
```
joshuagl:
```suggestion
provenance): do we remember 'good' property values, or do we remember 'good'
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+provenance? The difference is subtle.
+
+#### Alternative 1: Remember expected provenance values 
+
+With this alternative, the policy evaluator will compare selected properties of
+a package's SLSA provenance against the expected property values for that
+package. When a ‘good' package's provenance causes an unexpected difference,
```
joshuagl:
```suggestion
package. When a 'good' package's provenance causes an unexpected difference,
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+update the expected values.
+
+#### Alternative 2: Remember expected provenance instances
+
+Here, the policy evaluator will compare selected properties of a package's SLSA
+provenance against property values in past ‘good' provenance. When a ‘good'
```
joshuagl:
```suggestion
provenance against property values in past 'good' provenance. When a 'good'
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+package's provenance causes an unexpected difference, remember the new ‘good'
```
joshuagl:
```suggestion
package's provenance causes an unexpected difference, remember the new 'good'
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+provenance instance. 
+
+#### Expectations for multiple package versions
+
+A package dependency graph may require that multiple versions of the same
+package will be installed. For this reason, when expectations need to be updated
+for a package at a new version x.y.z (e.g., new source repository), then that
+update must not invalidate previous ‘good' versions of that package. The
```
joshuagl:
```suggestion
update must not invalidate previous 'good' versions of that package. The
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+solution is to qualify expectations for a package with a version range.
+
+-  With alternative 1,  different version number ranges correspond to
+    different expected provenance values .
+-  With alternative 2, different version number ranges correspond to
+    different past ‘good' provenance instances.
```
joshuagl:
```suggestion
    different past 'good' provenance instances.
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+#### Comparison of alternatives
+
+-  Both alternatives could work whether managed locally and centrally. 
+-  Both are compatible with the "trust on first use" approach described
+    below, because both generate provenance expectations, based on:
+    -  A configurable set of provenance properties that are not
+        expected to change. This could be an organization-wide or even
+        ecosystem-wide default setting.
+    -  The content of a ‘good' provenance instance.
```
joshuagl:
```suggestion
    -  The content of a 'good' provenance instance. 
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+provenance will or will not be compliant. Alternative 2 may be preferable if
+only information about the last package version needs to be remembered.
+
+## Bootstrapping policy support
+
+### Generating provenance expectations
+
+To overcome a bootstrapping problem, we assume that compromises will be rare,
+and propose to generate package-specific SLSA provenance expectations on first
+use. Each such ‘trust on first use' (TOFU) policy contains expectations based on
```
joshuagl:
```suggestion
use. Each such 'trust on first use' (TOFU) policy contains expectations based on
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+observed provenance, and can be used as a baseline to detect unexpected changes,
+for example a too low SLSA level, missing provenance, or an unexpected change in
+a ‘sensitive' provenance property. A [proposed TOFU
```
joshuagl:
```suggestion
a 'sensitive' provenance property. A [proposed TOFU
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+policy](#heading=h.56mzvgtcrr82) is presented below.
```
joshuagl: Is this a Google doc heading link? Should it be something like:
```suggestion
policy](#a-proposed-tofu-policy-for-a-single-npm-package) is presented below.
```
?

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+This is for a policy that lists explicit expectations for provenance values,
+instead of expectations for what provenance fields are invariant (regardless of
+their value).
+
+Inputs: 
+
+-  Package name and version.
+-  Optional: verified package provenance.
+-  Preferences. Examples include, but are not limited to:
+    -  Names of ‘sensitive' provenance properties that are not expected
```
joshuagl:
```suggestion
    -  Names of 'sensitive' provenance properties that are not expected
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+If a policy already exists for this package:
+
+-  Flag missing or unverifiable provenance.
+-  Flag an unexpected change in a sensitive provenance property (SLSA level,
+    source repo, etc.).
+
+The following will require approval:
+
+-  If a ‘good' package version does not satisfy expectations, add new
```
joshuagl:
```suggestion
-  If a 'good' package version does not satisfy expectations, add new
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+-  The policy applies only to the package that is being published (the
+    concrete versions of dependencies will be determined when a package is
+    downloaded for installation).
+
+####  Existing process details
+
+-  Besides package-specific code, an npm package contains a metadata file
+    (package.json) with the name and version of the package itself, and
+    optionally the names and allowed version ranges of direct dependencies.
+-  See "[Npm background](#heading=h.napkqzk41aim)" for more.
```
joshuagl: I don't think this background info section was copied over from the original Google doc?

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+The Appendix describes detailed scenarios to install packages and transitive
+dependencies. Despite differences in purpose, many are identical from a policy
+evaluation point of view:
+
+-  Installing all package dependencies of a project (Appendix
+    [A.1](#heading=h.vubk9ms4dtxh)).
+-  Adding a new package dependency to a project (Appendix
+    [A.2](#heading=h.vz2h8oar6gfd)).
+-  Installing an npm application locally (Appendix
+    [A.2](#heading=h.vz2h8oar6gfd)).
```
joshuagl: Are these all Google Doc heading links?

joshuagl (approved):
LGTM!

I left comments on minor nits like: left quotation marks that aren't straight quotation marks, url's that look like Google doc references, incorrect casing of npm, and other things. Fixing these is a little tedious but should help make the doc a more useful long-lived document.

0004/README.md
```diff
@@ -0,0 +1,514 @@
+# Policy Architecture for Npm Package Provenance
+
+*   Proposer: Wietse Venema ([wietse@gmail.com](mailto:wietse@gmail.com))
+*   GitHub Issue: TODO
+*   Status: [DRAFT](../README.md#meaning-of-status-codes)
+*   Original link: [Google docs](https://docs.google.com/document/d/1ulft22Nv6OXFeVRIrVfH2bK4QcRyvv0RwDt2dhJltK4/edit?usp=sharing) (see there for reviewer comments and rejected ideas)
+
+## Summary
+
+This document proposes an architecture for policy evaluation over SLSA
```
trishankatdatadog, line 10: Maybe I'm missing some context, but I see terms like "policy evaluation" and "policy" used early in this document, but it's not clear to me what they mean exactly. Perhaps a definition at the first time of use may be a good idea?

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+other lock file-based software environments such as Go (go.mod and go.sum), PHP
+(composer.lock), Python (Pipfile.lock), and Ruby (Gemfile.lock).
+
+## Objectives
+
+Evaluate policies over SLSA provenance for npm software packages.
+
+Non-objectives:
+
+-  Policies for what direct or indirect package dependencies (or dependency
```
trishankatdatadog, line 35: Yes, but this still unfortunately doesn't clarify exactly what policies are in the first place.

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+-  Other provenance claims may be relevant for specific use cases.
+
+When enforced, these expectations will ensure that a compromise of a package
+registry upload credential (or a compromise of the registry itself) cannot
+result in the installation of a rogue ‘foo' package version that was built from
+an unexpected repository or with an unexpected builder. 
+
+If a ‘good' package version does not meet policy expectations, the expectations
+need to be updated for that package version and onwards, so that earlier package
+versions will continue to pass expectations (differnt dependencies may require
```
trishankatdatadog:
```suggestion
versions will continue to pass expectations (different dependencies may require
```

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+different versions of a package).
```
trishankatdatadog, line 100: Hmm, this raises an interesting Q (which I have run into practice): how does one know which policy versions to apply to different package versions?

0004/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+These expectations cannot detect 'upstream' compromises, for example the
```
trishankatdatadog: Sorry, why not exactly? It's not entirely clear to me right now, especially if you tie the right types of attestations together, combined with a sufficiently expressive policy.

trishankatdatadog (commented):
Haven't read the whole proposal, unfortunately, but attached some high-level comments so far that I hope will help FWIW.

0004/README.md
kommendorkapten: With the policies mentioned here, are mostly around verifying that the correct upstream repository and builder was used. There are some loose mentions of other claims too. But in the general case I think it would be hard to detect the examples listed below with just a SLSA build provenance? 
* If the builder/build system is compromised that would require deep introspection of the system, e.g. trace syscalls to see that a file is being replaced during build/packaging time etc.
* Malicious change to a source repository. This would be impossible(?) to detect during build, and would require source level attestations?

0004/README.md
kommendorkapten: Yes, npm is always lower case: https://github.com/npm/cli#is-it-npm-or-npm-or-npm

kommendorkapten (approved):
LGTM!

Some formatting issues only that @joshuagl already pointed out.

0004/README.md
wietse-gmail: Sorry that should be a link to the "npm background"  section  in *this* document. The link was not updated when the Google doc was converted to markdown. I'll update the README.md.

0004/README.md
wietse-gmail: These were not updated by the Gdoc-to-Markdown conversion. I'll update these links too.

0004/README.md
wietse-gmail: As written, one policy must be able to cover multiple package version ranges. This could be implemented with rules of the form (version-range, expectations), where version-range is a semantic version pattern.

0004/README.md
wietse-gmail: Would it help if I add some text along the lines of: "A provenance policy associates a package name with expectations for provenance attestation properties. A package (name, version) satisfies the provenance policy iff its provenance meets the expectations for that package name."

0004/README.md
wietse-gmail: Would it help if I add text along the lines of: "A provenance policy specifies expectations for provenance attestation properties. It can protect only the part of the supply chain after the attestation is signed. Not the parts that come before: the build process, or the source repository content."

0004/README.md
trishankatdatadog: That would go a long way, thanks, although I also do think it helps to add a small `Background` or `Architecture` section or something similar later on that goes deeper these definitions.

0004/README.md
wietse-gmail: I updated the quotes as suggested and replaced all in-document links.

0004/README.md
wietse-gmail: I have added a terminology section before "Objective" that covers "provenance policy" and what "satisfying a policy" means. Elsewhere there already is text that covers expectations and attestation properties at a high level, but I'm not sure this document is the right place for deeper details of how policy evaluation is implemented.

0004/README.md
wietse-gmail: This concern duplicates an earlier comment. I'll resolve this one, so that we don't dilute the discussion.

0004/README.md
wietse-gmail: I updated the text so that it  now covers both a) the part of the software supply chain that the  expectations can protect, and b) the part that they cannot. protect

wietse-gmail:
@trishankatdatadog I added clarifications for the issues that you raised. Will this work for you?
Unfortunately I haven't figured how to view these comments AND the updated text together, that would be a time saver.

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+With few differences, the approach presented here is expected to generalize to
+other lock file-based software environments such as Go (go.mod and go.sum), PHP
+(composer.lock), Python (Pipfile.lock), and Ruby (Gemfile.lock).
+
+## Terminology
+
+*   A *provenance policy* associates a package name with expectations for the provenance attestation properties for a collection of package versions. 
+
+*   A package version *satisfies* a provenance policy iff its provenance attestation properties meet the policy expectations for the package name.
```
feelepxyz:
```suggestion
*   A package version *satisfies* a provenance policy if its provenance attestation properties meet the policy expectations for the package name.
```

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+-  Preferences. Examples include, but are not limited to:
+    -  Names of 'sensitive' provenance properties that are not expected
+        to change such as builder ID or source repository,
+    -  Minimum SLSA level for an acceptable artifact, 
+    -  Maximum SLSA level depending on the builder. 
+
+-  Preferences may differ for different evaluator contexts: developer,
+    registry, or end user, and for a private or public registry.
+
+If no policy exists for this package, generate one with:
```
feelepxyz, line 303: @wietse-gmail where you thinking we would create a separate policies file in the current directory or place this information in the lockfile somehow? Probably out of scope for this doc but curious to think about some specifics to see if I'm following here.

It also seems like we would need the ability to update this policy
- First time you see a version with provenance, create a policy entry for that package name and `>= version`
- If you later happen to install a lower version of the same package that also has provenance, update the policy version pattern to include this lower version

This would also mean the policy we can create on the registry would not necessarily be the same on you get locally when installing. Not sure this is a major problem though? Keeping a canonical policy on the registry also seems like a lot of work. Pretty sure we wouldn't want to stuff it into the [package metadata object](https://registry.npmjs.org/sigstore). 

We have been thinking of exposing a list of trusted builders from the registry that the CLI could use when verifying.

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+This is for a policy that lists explicit expectations for provenance values,
+instead of expectations for what provenance fields are invariant (regardless of
+their value).
+
+Inputs: 
+
+-  Package name and version.
+-  Optional: verified package provenance.
+-  Preferences. Examples include, but are not limited to:
```
feelepxyz, line 294: Specifically, I guess this would include things like
- repository url
- builder id (without version suffix?)
- repository id
- owner id
- workflow ref, job workflow ref (without the branch name?)

Seems like we wouldn't be able to match on tag or branch name as this might cause a lot of false positives when people use release branches. Event name might also be a problem if you legitimately use two events to publish.

0004/README.md
wietse-gmail: Sorry for the late response; I was on vacation and got busy catching up.

For the initial phase I was thinking of storing auto-generated policy expectations locally, separate from the package-related files that already exist (no invasive change).

An auto-generated policy can indeed differ between developer, registry, and end user; the result reflects what package versions have been evaluated. Would there be other differences?

I agree that it makes sense to have a separate policy for what builders etc. to trust. This can be site-specific, like the preferences that I mention a little later.

0004/README.md
wietse-gmail: Yes, there needs to be a way to constrain what claims provenance can make, with some granularity that is still TBD. A possible action could be to downgrade a package to a low SLSA level or to reject it.

0004/README.md
wietse-gmail: @trishankatdatadog Are you OK with the text as amended?

0004/README.md
wietse-gmail: @trishankatdatadog Are you OK with the text as updated?

0004/README.md
wietse-gmail: @feelepxyz Philip, is this clarification sufficient or should I expand the text with these implementation details?

0004/README.md
wietse-gmail: @feelepxyz Does this address your concern or should I add text that looks ahead into implementation?

0004/README.md
trishankatdatadog: Sorry, been super busy, let me try to find some downtime...

0004/README.md
wietse-gmail: @fridex I added a note in the References section (https://pypi.org/project/pipfile/0.0.2/ covers lock files for Python but is not yet official). Does that address your concern?

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+With few differences, the approach presented here is expected to generalize to
+other lock file-based software environments such as Go (go.mod and go.sum), PHP
+(composer.lock), Python (Pipfile.lock), and Ruby (Gemfile.lock).
+
+## Terminology
+
+*   A *provenance policy* associates a package name with expectations for the provenance attestation properties for a collection of package versions. 
+
+*   A package version *satisfies* a provenance policy if its provenance attestation properties meet the policy expectations for the package name.
```
trishankatdatadog: Maybe more accurate to say "a version of a package"

trishankatdatadog:
```suggestion
*   A package version *satisfies* a provenance policy if its provenance attestation properties meet the policy expectations for the package name (and, naturally, version).
```

MarkLodato:
To get this submitted, let's set a deadline for end of Thursday (May 11). If there no more significant comments, we'll merge.

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+![Extended model](extended-model.svg)
+
+The figure shows a generic supply chain for software packages, from source repository
+to installation, extended with two SLSA provenance policy evaluation points:
+
+-  When a software package is published to a package registry.
+-  When a software package is installed from a package registry.
+
+A SLSA provenance policy contains expectations for provenance, as discussed in later
```
trishankatdatadog, line 61: Sorry, it's a bit hard to understand what "expectations" are w/o first defining them upfront. The next paragraph doesn't make sense to me until I reread it again later.

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+### SLSA provenance expectations, benefits and limitations
+
+Each SLSA provenance policy covers multiple versions for a specific package
+name. For example, the policy for package ‘foo' could expect that each covered
+version:
+
+-  Has SLSA provenance that is signed with a key that is bound to a trusted
+    identity. The key could be long-lived, rotated regularly, or ephemeral, and
+    the identity could be a builder, individual, or organization.
+-  The provenance contains claims that:
```
trishankatdatadog, line 81: What are these claims? Do they come from one or more attestations? If more, are they different types of attestations beyond just (build) provenance?

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+    -  Combined together satisfy the requirements for SLSA level X.
+        -  A policy evaluator may limit the maximum SLSA level
+            depending on the builder identity or other context. 
+
+    -  The package was built from the repository
```
trishankatdatadog: Do you mean an SCM repository?

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+        [https://example.com/path/to/foo](https://example.com/path/to/foo).
+
+-  Other provenance claims may be relevant for specific use cases.
```
trishankatdatadog: Again, what are these claims? Also, would you please elaborate on some exemplary use cases?

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+When enforced, these expectations will ensure that a compromise of a package
+registry upload credential (or a compromise of the registry itself) cannot
+result in the installation of a rogue ‘foo' package version that was built from
+an unexpected repository or with an unexpected builder. 
+
+If a ‘good' package version does not meet policy expectations, the expectations
```
trishankatdatadog, line 97: Sorry, but how? Because now you have two problems instead of one: updating packages _and_ policies securely. Shouldn't a policy itself be versioned? If so, wouldn't you associate different policies for different ranges of versions of packages?

0004/README.md
```diff
@@ -0,0 +1,524 @@
...
+Initially the policy evaluator will run post-installation, producing alerts
+only, using expectations stored at the local evaluator instance. For enforcement
+the evaluator will need to be integrated into the package installer, and
+expectations may become managed centrally. 
+
+### Generating provenance expectations
+
+To overcome a bootstrapping problem, we assume that compromises will be rare,
+and propose to generate package-specific SLSA provenance expectations on first
+use. Each such ‘trust on first use' policy contains expectations based on
```
trishankatdatadog, line 150: I'm not sure TOFU policies will scale beyond a certain point. How should package developers, package registries, and package users be expected to keep updating their policies when things break?

trishankatdatadog (commented):
First, thanks very much for writing this, Wietse, and for pushing this, Mark, I really appreciate the efforts. However, I am not sure that SLSA is ready for policies just yet. For one thing, I strongly recommend that we use [in-toto layouts with policy definitions](https://github.com/in-toto/ITE/pull/38) to solve the problem in a standardised manner. For another, it's still unclear to me how layouts (containing policies) will be distributed---and, more importantly, updated---securely, but to me that is a problem we can worry about later on. For now, I think it's important that, as we standardise more and more in-toto attestations, we finally introduce the full power of in-toto layouts (containing policies), perhaps in a SLSA 2.0 instead of reinventing the wheel. Does this make sense? Please feel free to disagree, and I look forward to discussing and pushing this forward.

0004/README.md
wietse-gmail: Updated.

wietse-gmail: Adopted with one change: a policy applies to package version *ranges*, not individual versions. That is, versions foo-a.b.c .. foo-x.y.z may have different expectations than earlier or later versions. @MarkLodato FYI.

0004/README.md
wietse-gmail: This text considers only *provenance* attestations for a specific package version artifact. If I suggest otherwise then I should fix that.

There may be information to limit  trust  based on builder ID or other context.  Those details are not covered here.

0004/README.md
wietse-gmail: Changed to 'source repository' here and elsewhere in the doc.

0004/README.md
wietse-gmail: Added a prefix with "out of scope", and added an example: local extensions for internal use.

0004/README.md
wietse-gmail: There is policy and there is distribution. Without policy there is no distribution, and this doc is primarily concerned with policy.

Provenance policy expectations need to support package versions (ranges). This is now explicit in 'terminology' and a few other places in the text. I can add more if desirable.

The TOFU approach is for bootstrapping in the absence of a secure policy management and distribution system.  It allows us to learn what works, policy-wise.

0004/README.md
wietse-gmail: Indeed, TOFU is only a for bootstrapping, not a long-term, full-scale solution. If I need to make that more explicit, then I will do that.

0004/README.md
wietse-gmail: I added some examples here, in the terminology section, and below in the section on 'SLSA provenance expectations'. Does that help?

trishankatdatadog: Yes, thanks!

0004/README.md
trishankatdatadog: Makes sense for prototyping purposes, as @MarkLodato explained earlier today. Could we please mark distribution as out of scope for this proposal?

0004/README.md
trishankatdatadog: SGTM. Following my earlier comment, could we please mark distribution as out of scope for this proposal, and explicitly state that TOFU is not a long-term solution? Thanks!

trishankatdatadog (commented):
Thank you very much for your hard work, Wietse!

I am happy to approve this PR so long as we document that:
1. We are not yet defining exactly what _policies_ must look like, and how they must be presented. We are experimenting with them right now, including natural options such as [in-toto layouts with policies](https://github.com/in-toto/ITE/pull/38).
2. We are not prescribing how _distribution_ of policies must look like. We are beginning our experiments with TOFU, but may not end up using them in the long term.

Does this sound good to everyone? @MarkLodato @adityasaky

0004/README.md
wietse-gmail: Added in the non-objectives section under  'policy distribution'

0004/README.md
wietse-gmail: Added under  'non-objectives/policy distribution' that the TOFU approach supports prototyping policy and policy evaluation, and that it is not an alternative for secure management and distribution.

0004/README.md
```diff
@@ -34,6 +34,12 @@ Non-objectives:
     versions) to expect for an npm software package; policies for vulnerability
     scans.
 
+-  Policy distribution. The TOFU (trust on first use) approach described here supports
+    protyping policies and evaluators, and is not an alternative for secure management
+    and distribution.
+
+-  Prescribe policy formats. The goal is to prototype policies and evaluators.
```
joshuagl: This last sentence ("The goal") would work well as an explicit objective.

wietse-gmail: Done. Great idea :-)

MarkLodato:
> 1. We are not yet defining exactly what _policies_ must look like, and how they must be presented. We are experimenting with them right now, including natural options such as [in-toto layouts with policies](https://github.com/in-toto/ITE/pull/38).
> 2. We are not prescribing how _distribution_ of policies must look like. We are beginning our experiments with TOFU, but may not end up using them in the long term.

Sounds good to me.

I think it makes sense to submit this now as DRAFT with the clear intention of us iterating as we learn more.

joshuagl:
> I think it makes sense to submit this now as DRAFT with the clear intention of us iterating as we learn more.

Completely agree.

joshuagl (approved):
Thanks for working through everyone's questions @wietse-gmail !

LGTM to merge as draft and iterate

MarkLodato:
Sounds great. @trishankatdatadog could you approve and merge if you agree?

trishankatdatadog (approved):

wietse-gmail:
Thank you all!

PR #9, opened 2023-04-28
SLSA Conformance Program

kpk47:
## Abstract

This document describes a mechanism for labeling build platforms as officially
SLSA conformant, the "SLSA Conformance Program." This proposal lays out the
program at a high level, including a draft of the survey build platforms would
need to fill out to participate. Please focus review on the program as a whole
and the survey's general contents, reserving discussion of the survey's final
wording for implementation PRs.


0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+The OpenSSF would maintain a repository that stored all public proof of
+conformance and provenance verification materials for build platforms taking
+part in the program. This option made enforcement easy since you could remove
+build platforms that do not comply with the program's terms, but it puts too
+much trust in the repository's maintainers. This option would also require
+infrastructure for build platforms to rotate and revoke provenance verification
+materials, which would complicate the design.
+
+### Build platform attestations
```
mlieberman85: You could still provide it as a separate attestation document entirely and one that doesn't necessarily need to be updated all the time. Even if we go the human readable route, I do wonder if folks are going to want some machine readable content to help with some of the work or for internal consumption for compliance.

e.g. the same human readable content just as key/values in a JSON that could then be ingested into an organizations risk management/governance/compliance platform.

mlieberman85:
I like what's here so far. I think there's some questions and concerns though.

How does this discern between build software and configuration, e.g. Tekton, Jenkins, etc. and running build tools? I know SLSA conformance is about running build systems, but we need to figure out a way to ensure that it's not a huge burden for folks to prove conformance when they run their own build system. There's currently a misconception that SLSA pushes for SaaS over self-hosted. We need to think of a way to make that easy.

0004/README.md
kpk47: How about we require that proof of conformance docs and audit reports be available in a computer-readable format? That way they're easy to manipulate, but we aren't passing them around in the provenance.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+this principle's asymmetry through how it treats [verifying build
+platforms](https://slsa.dev/spec/v1.0/verifying-systems) and [verifying
+artifacts](https://slsa.dev/spec/v1.0/verifying-artifacts) – the former is a
+list of freeform questions a verifier might wish to ask of a build system, and
+the latter is a detailed procedure to be implemented in a tool. It is difficult
+and expensive to answer the question "is this build platform SLSA conformant?"  
+
+As a practical matter, the OpenSSF controls SLSA and could be in a position to
+determine which build platforms are SLSA conformant. However, the OpenSSF
+doesn't have the desire or resources to do so. Furthermore, the SLSA
```
arewm: Maybe combine these first two sentences?
```suggestion
While the OpenSSF controls SLSA and could therefore be in a position to
determine which build platforms are SLSA conformant, it
doesn't have the desire or resources to do so. Furthermore, the SLSA
```

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+specification makes it the verifier's responsibility to determine whether a
+build system is trustworthy, and that trustworthiness determines Build L3
+conformance. Instead, the OpenSSF should do what it does best: build a
+community. In this case, that community would consist of build platforms,
+verifiers, and independent auditors who hold each other accountable for build
+platform security.   
+
+The OpenSSF's tool for building this community is a SLSA Conformance Program.
```
arewm: This wording feels awkward to me. I think it is a combination of the leading `The` as well as introducing the conformance program as a tool. Is `The OpenSSF` the proper way to refer to OpenSSF or can it be shortened to just OpenSSF?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+CNCF's Certified Kubernetes Software Conformance -
+[https://www.cncf.io/certification/software-conformance/](https://www.cncf.io/certification/software-conformance/)
+
+## Proposed changes
+
+### Overview
+
+The OpenSSF maintains a SLSA Conformance badge and registry of build platforms
+and auditors that take part in the SLSA Conformance Program. Participating in
```
arewm: Are there multiple `and`s in the sentence because they are not part of the same series? If they are not part of the same series, can we use different conjunctions so that readers don't try to merge into one list?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+Rather than maintain offline proof of conformance that would be written and
+interpreted by humans, this option would capture the state of the build platform
+in an attestation and attach it to all provenance produced by the build
+platform. We discarded this option because it would violate SLSA's guiding
+principle of trusting systems by moving conformance checks into provenance
+verification. It would also bloat the provenance size and require introducing
+new, complicated logic in the provenance verifier.
+
+## Appendix A: Conformance questionnaire
```
arewm: How does this relate to https://slsa.dev/spec/v1.0/verifying-systems?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+#### Tier one: Self-attestation
+
+The self-attestation tier requires that a build platform:
+
+-  Publish answers to the questionnaire on a domain registered to their
+    organization. They may additionally publish further proof of SLSA
+    conformance (e.g. architecture documentation). This information must be
+    updated at least once every twelve months and any time there is an
+    architectural change to the build platform.
```
arewm: What information? The questionnaire answers? Architecture documentation? Both? If there are no _drastic_ changes, would updating the date constitute an update or would a more noticeable change be required?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+-  Publish any materials needed to verify provenance signed by the build
+    system on a domain registered to their organization.
```
arewm: Is this something like a tool capable of creating a [VSA](https://slsa.dev/verification_summary/v1) in an automated/scripted fashion?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+-  Publish vulnerability disclosures for all security incidents that
+    potentially impact the integrity of artifacts or provenance generated by
+    the build system.
```
arewm: Is this the vulnerabilities affecting the artifacts itself or that affect the build system and therefore the artifacts that were build on the system? By the intention of the program, I assume the latter, but due to the overhead of the former, I had to ask.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+The OpenSSF operates this tier through a community-maintained repository of
+build platforms. The maintainers of this repository will be drawn from members
+of the SLSA community who are spread across the industry.  
+In order to participate in this tier, build platforms should 
+
+1. Learn about the SLSA Specification and its requirements.
+1. Create a PR for the repository containing
+    1. Name of the build platform
+    1. Build ID
+    1. Desired SLSA levels
```
arewm: According to v1.0, the build ID should map uniquely to a SLSA level. I have a question about this as it relates to higher SLSA levels: https://github.com/slsa-framework/slsa/issues/849. Based on this, I feel like we are starting to think that it would be _okay_ to loosen the 1:1 requirement.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+    potentially impact the integrity of artifacts or provenance generated by
+    the build system.
+-  Sign and follow the SLSA Conformance Program terms.
+
+The OpenSSF operates this tier through a community-maintained repository of
+build platforms. The maintainers of this repository will be drawn from members
+of the SLSA community who are spread across the industry.  
+In order to participate in this tier, build platforms should 
+
+1. Learn about the SLSA Specification and its requirements.
```
arewm: Nit: Are all of these supposed to be `1`'s?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+1. Create a PR for the repository containing
+    1. Name of the build platform
+    1. Build ID
+    1. Desired SLSA levels
+    1. List of URIs that resolve to proof of SLSA conformance
+    1. List of URIs that resolve to materials for verifying provenance
+    1. List of URIs that resolve to the build platform's public
+        vulnerability disclosures
```
arewm: Are these URIs the `builder.id` or are you trying to associate some other URI with these references?

I think we can simplify this part by grouping all together under a common, something like
> List of URIs that resolve to the self-reported evidence

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+1. Complete the terms and conditions form.
+1. Send the PR for review.
+
+The community will review the PR and merge it when a quorum of maintainers
```
arewm: Can you link to the list of maintainers that would vote on it?

arewm (commented):
Initial comments. I will need to re-review starting with third-party audit.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+    architectural change to the build platform.
+-  Publish any materials needed to verify provenance signed by the build
+    system on a domain registered to their organization.
+-  Publish vulnerability disclosures for all security incidents that
+    potentially impact the integrity of artifacts or provenance generated by
+    the build system.
+-  Sign and follow the SLSA Conformance Program terms.
+
+The OpenSSF operates this tier through a community-maintained repository of
+build platforms. The maintainers of this repository will be drawn from members
```
chtiangg: To clarify, the maintainers of this repo, the community, and the verifier (mentioned in the "Problem statement" section) are the same group of people. Would it mean that a verifier has to be the repo maintainer, which leads to another question: what is the criteria to become a maintainer of this repo? Or what is the requirement for "Verifier Participation"?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+of the SLSA community who are spread across the industry.  
+In order to participate in this tier, build platforms should 
+
+1. Learn about the SLSA Specification and its requirements.
+1. Create a PR for the repository containing
+    1. Name of the build platform
+    1. Build ID
```
chtiangg: I ask this question may because I am new to the open source world:

How would people ensure the Build ID must be only used by the self-attested build platform? If I am a random builder, but directly use that BuildID to produce my provenance, is there anything to prevent it or to say I cannot use this BuildID? (I guess signature on the provenance is the traditional way, but this "Tier one" doesn't require registering a public key in the PR.)

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+    potentially impact the integrity of artifacts or provenance generated by
+    the build system.
+
+This tier's requirements are enforced by the auditor.   
+The OpenSSF will run a community-maintained repository of approved auditors and
+audited build platforms. This repository may be the same one used for the
+self-attestation tier.  
+In order to participate in this tier, build platforms should
+
+1. Successfully complete an audit with an approved auditor
```
laurentsimon: I wonder whether we should recommend providing a reference implementation, possibly in the slsa-verifier. This encourages a single codebase which would benefit users. It would also encourage builders to think about key management (that's a core design principle of the slsa-verifier today that verifiers should not be required to manage keys), which aligns with the "root of [builder] trust" in SLSA.

0004/README.md
```diff
@@ -0,0 +1,416 @@
+# Proposal 4: SLSA Conformance Program
+* Proposers: [Kris K](https://github.com/kpk47) (kkris@google.com), [Joshua Mulliken](https://github.com/JoshuaMulliken) (jmullike@redhat.com)
+* GitHub issue: https://github.com/slsa-framework/slsa/issues/515
+* Status: [DRAFT](../README.md#meaning-of-status-codes)
+
+## Abstract
+
+This document describes a mechanism for labeling build platforms as officially
+SLSA conformant, the "SLSA Conformance Program." This proposal lays out the
+program at a high level, including a draft of the survey build platforms would
```
trishankatdatadog:
```suggestion
program at a high level, including a draft of the survey that build platforms would
```

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+One of SLSA's
+[guiding principles](https://slsa.dev/spec/v1.0/principles#trust-platforms-verify-artifacts)
+is to "Trust platforms, verify artifacts." The SLSA v1.0 specification reflects
+this principle's asymmetry through how it treats [verifying build
+platforms](https://slsa.dev/spec/v1.0/verifying-systems) and [verifying
+artifacts](https://slsa.dev/spec/v1.0/verifying-artifacts) – the former is a
+list of freeform questions a verifier might wish to ask of a build system, and
+the latter is a detailed procedure to be implemented in a tool. It is difficult
+and expensive to answer the question "is this build platform SLSA conformant?"
```
trishankatdatadog: Please briefly explain when this may or may not be the case. For example, anyone can easily answer the question whether a build platform produces any signed/unsigned provenance, but not whether it is isolated and hosted.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+1. Set a baseline for Build L3 build platforms
+1. Bootstrap an ecosystem/market of build platforms and auditors
+1. Create a sensible process for managing SLSA trademarks
+1. Limit or mitigate perverse incentives for auditors
+1. Scale from new market entrants to existing large providers
+
+Nice to have:
+
+1. Distribute provenance verification materials
```
trishankatdatadog: How would this alone answer whether a build platform is v1.0 L3-compliant? If it cannot, why is this here?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+build platforms. The maintainers of this repository will be drawn from members
+of the SLSA community who are spread across the industry.  
+In order to participate in this tier, build platforms should 
+
+1. Learn about the SLSA Specification and its requirements.
+1. Create a PR for the repository containing
+    1. Name of the build platform
+    1. Build ID
+    1. Desired SLSA levels
+    1. List of URIs that resolve to proof of SLSA conformance
```
trishankatdatadog: I think "evidence" is a better word than "proof," the latter of which is usually reserved for mathematically provable statements.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+information in the proof of conformance is truthful, and they will approve it if
+the described build platform meets the requirements for the desired SLSA levels.
+
+Once the PR is merged and the terms are signed, the OpenSSF grants the build
+platform a limited license to use the SLSA logo on its website and other
+promotional materials. Should the build platform fail to adhere to the program's
+terms or meet its requirements, the OpenSSF will revoke that license and, if
+necessary, enforce its trademarks. 
+
+#### Tier two: Third-party audit
```
trishankatdatadog: I propose we actually create a new in-toto [predicate](https://github.com/in-toto/attestation/tree/main/spec/predicates) which can be used by third-party auditors to sign new in-toto attestations to the effect that they vouch for the given level of SLSA compliance for a build platform. I see that a build platform attestation has been proposed and rejected below, the latter of which I believe is too early to do so.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+materials, which would complicate the design.
+
+### Build platform attestations
+
+Rather than maintain offline proof of conformance that would be written and
+interpreted by humans, this option would capture the state of the build platform
+in an attestation and attach it to all provenance produced by the build
+platform. We discarded this option because it would violate SLSA's guiding
+principle of trusting systems by moving conformance checks into provenance
+verification. It would also bloat the provenance size and require introducing
```
trishankatdatadog: I disagree with these particular reservations, because:
1. We shouldn't conflate the build _provenance_ with any new build _platform_ attestation.
2. New in-toto predicates will come in the future, thus necessarily increasing the total size of any bundle of attestations.
3. (2) will necessarily add new logic to any SLSA verifier.

0004/README.md
arewm: Upon re-reading, I am not sure what this actually entails. Is it a distribution of tools to help verify the provenance (and therefore SLSA build levels) for various platforms? Is it just the public keys needed for provenance publication?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+-  Provide the auditor access to the build system and any supporting
+    materials needed for the auditor to gain trust in the build system.
+-  Accept that the audit report will be made public regardless of the
+    audit's outcome. Audit reports may be delayed up to 90 days to respond to
+    vulnerabilities in the build platform, but they must be published within 30
+    days of patching the vulnerability.
+-  Provide verifiers with any materials needed to verify provenance signed
+    by the build system. Publishing these materials on a domain registered to
+    the build system meets this requirement.
```
arewm: Similar to the earlier comment, it isn't clear what is intended by this. If there is some corollary in the SLSA specification, can we link to it to complete the reference?

0004/README.md
arewm: Are there any requirements about publicity of notices or how quickly the notices should be published after notification/patching?

The questions are relevant for third-party attestation as well.

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+-  Notify known build platform users of any security incidents that
+    potentially impact the integrity of artifacts or provenance generated by
+    the build system.
+
+This tier's requirements are enforced by the auditor.   
+The OpenSSF will run a community-maintained repository of approved auditors and
```
arewm: I assume approved auditors are just those which pass the below criteria. Can we link to that criteria since it is introduced later? Or introduce the concept earlier in the document?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+self-attestation tier.  
+In order to participate in this tier, build platforms should
+
+1. Successfully complete an audit with an approved auditor
+1. Create a PR for the repository containing
+    1. Build platform name
+    1. Audited SLSA levels
+    1. URI that resolves to the public audit report
+    1. (Optional) Build ID
+    1. (Optional) List of URIs that resolve to proof of SLSA conformance
```
arewm: How is this different than the public audit report?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+    1. (Optional) List of URIs that resolve to materials for verifying
+        provenance
+    1. (Optional) List of URIs that resolve to the build platform's
+        public vulnerability disclosures
```
arewm: What is the rationale behind these being optional here but not on the self-attested conformance?

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+1. Complete the terms and conditions form.
+1. Send the PR for review.
+1. Have the auditor approve the PR with a comment indicating the audited
+    SLSA levels
```
arewm: Why is the auditor not creating the PR? Is it so that an identity is associated with the build platform which the auditor just then needs to approve?

It seems like the auditor should create the PR as they are the one submitting it. One challenge for this would be how would approvers know a sufficient identity of the build platform to approve (if needed).

0004/README.md
```diff
@@ -0,0 +1,416 @@
...
+unique identifiers for each build's output artifacts.
+### Continuous Monitoring and Improvement
+
+To ensure ongoing compliance with the SLSA Build L3 requirements, builders
+should have a continuous monitoring and improvement plan in place. This plan
+should include the following elements:
+
+1. **Regular Security Reviews**: Conduct periodic reviews of security
+    practices, threat models, and risk assessments to identify areas for
+    improvement and ensure that the builder's security posture remains up-to-date.
```
arewm: Are there any considerations on entities that should be used to perform the reviews? Is an internal review acceptable? Do these need to be published?

0004/README.md
arewm: I guess that might relate to this in the `Demonstrating compliance` section below?

0004/README.md
kpk47: Done

0004/README.md
kpk47: Their website uses both versions, so I'll drop the 'the's.

0004/README.md
kpk47: I meant public certificates or sigstore metadata. I've updated the text to clarify.

The provenance signing materials link provenance to a build platform. It isn't strictly necessary to store the verification materials alongside the evidence of conformance, but it seems like a logical place to put it. If you feel strongly, I can remove it.

0004/README.md
kpk47: Done

0004/README.md
kpk47: Done

0004/README.md
kpk47: No, this is verification keys. I've clarified here and below.

0004/README.md
kpk47: Done

0004/README.md
kpk47: No. The maintainers of the conformance repo are a subset of the SLSA community. The verifier is defined in the SLSA spec: https://slsa.dev/spec/v1.0/terminology#roles.

I don't know how exactly we'll pick the maintainers for the conformance repo, but realistically we'll probably take volunteers from the SLSA community.

There is no barrier to verifier participation. Anyone who has both the provenance and its verification key can verify it.

0004/README.md
kpk47: No, that's an artifact from the tool I used to convert this PR from a Google doc. Fixed.

0004/README.md
kpk47: You're right -- you need to use the signature on the provenance envelope layer to be sure that the build ID is used correctly. I've clarified that "verification materials" are public keys.

0004/README.md
kpk47: That wasn't my intention. Consider the s to be a typo.

That being said, if a platform is capable of producing different level provenance based on user inputs then we have to trust it to use the right builder ID. The process they use to choose the builder ID could also produce a VSA, but I'm not convinced it buys us anything. I'm not decided yet.

0004/README.md
kpk47: Done.

0004/README.md
kpk47: I was thinking other URIs, probably URLs on the build platform's website.

I can't tell from your comment which entries you'd like consolidated. Can you clarify?

0004/README.md
kpk47: I don't have a list and am not entirely sure the best way to get one. I was thinking maybe volunteers from the current SLSA contributors with some limit to the number who can be at one company, but I'm open to suggestions.

0004/README.md
kpk47: What would go in this attestation, and who is verifying the signature? Is the idea to verify the attestation as part of provenance verification, or would you do it once as part of adding a trust root to a verifier tool?

0004/README.md
kpk47: The idea is to share verification keys along with proof of conformance. Doing so lowers the barrier to verification since you can query the registry to determine whether you trust a build platform and get its public key.

I must admit that I've also been thinking about how slsa-verifier can get additional trust roots without having to hard code them, so I may be conflating the two issues in my head.

0004/README.md
kpk47: Moved auditors first.

0004/README.md
kpk47: Do you mean a reference implementation of an audit? I'd thought slsa-verifier would draw from the conformance registry to get new trust roots, not be an auditor. Or am I misunderstanding you?

0004/README.md
kpk47: It should read "supplemental evidence," though maybe that should be included in the audit report. I'll drop it.

0004/README.md
kpk47: I made them mandatory.

0004/README.md
kpk47: I've changed it so that the auditor sends the PR.

There's no need for an approval from the build platform because the audit report must always be made public and requiring an approval would allow the build platform to suppress the report.

0004/README.md
kpk47: It's derived from that page. I added a note explaining that connection.

0004/README.md
kpk47: I think internal-only is fine.

0004/README.md
laurentsimon: I added the comment probably at the wrong location. I agree with what you say. I meant reference implementation for the verification of provenance generated by the builder. Builders don't just need a key to be verified. They often come with a specific buildType, e.g. GCB provenance != GH provenance != GitLab provenance. This logic needs explanation and implementation. See https://slsa.dev/provenance/v1 `Index of build types` section.

Lmk if this makes no sense :)

0004/README.md
arewm: I was thinking about modifying the list to put all under a common heading. Something like this (i.e. to remove the redundancy around "list of URIs that resolve to"):

* List of URIs that resolve to the self-reported evidence
  * Proof of SLSA conformance
  * Materials for verifying provenance
  * Build platform's public vulnerability disclosures

0004/README.md
arewm: Would this be considered everyone with write access to the repository? What is considered a quorum? Is it a simple majority? Is there a minimum number that need to be approving? Some specific subset of maintainers?

> Quorum:
> the minimum number of officers or members of a body that is required to be present at a given meeting (as to transact business)

0004/README.md
kpk47: I see what you mean. I've added a requirement that build platforms provide a reference verification implementation and suggested it be contributed to slsa-verifier.

0004/README.md
kpk47: We won't really know how many maintainers there are until we're able to gather volunteers, so let's go with 2+. We can always change it if that doesn't work.

0004/README.md
kpk47: I think we're talking about different ideas. I was thinking about attestations about the machine state that would be validated against a SLSA conformance policy or test suite, but you were suggesting attestations for auditors to attest to SLSA levels.

I'm against attestations about machine state because I don't think it's possible to write a policy or test that describes a sufficiently secure build platform. I'm aware of some efforts to do so, but I don't think they're widely accepted enough to be part of SLSA.

I don't have strong feelings about auditor attestations in general. However, I don't think they're necessary to start a conformance program because we could easily add them later on top of the registry. We would also need to engage with the complexity of how to use those attestations (e.g. at each verification, once when configuring the verifier's root of trust?), and I would prefer to start with a simple approach and add complexity as needed.

kpk47:
Thank you, everyone, for the discussion. I've let this PR sit open for too long, and I would like to merge it in DRAFT state on June 30, 2023. Please add any additional comments by then, and I will try to address them. We can address any comments after that date with new PRs.

0004/README.md
trishankatdatadog: Good Qs. I imagine it's something you download, cache, and verify every now and then (although it does need to come with an expiration for obvious reasons).

PR #10, opened 2023-05-17
Adds the proposal for restructuring the versioning system for SLSA

chtiangg:
This PR added the draft version of the proposal for review.

Even though the google doc has been reviewed from a few repo maintainers, still marked as DRAFT to gain additional feedback from the community if any.


0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+The following is a snapshot of the scripts for building and deploying the
+website.
+
+-   To serve the website locally,
+    [/docs/README.md](https://github.com/slsa-framework/slsa/blob/main/docs/README.md)
+    is replaced with running `build.sh` or similar logic.
+-   The Github Workflow uses the same shell script to prepare the build
+    environment.
+    
+```
```
MarkLodato: nit: I'm not sure we really need a script here. The only difference from the existing instructions is how to checkout (using submodules). Otherwise it's identical, right? That would be good to explain, to show how it's not a big change.

I have comments about the github config, but I'll put that as a separate comment.

(If you keep it, use ```` ```bash ```` to enable syntax highlighting.)

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+bundle install
+
+if [[ "${serve_website}" == "true" ]]; then
+  echo "Start to serve the website..."
+  bundle exec jekyll serve --livereload --incremental
+else
+  echo "Finish setting up code and installing dependencies. Exit."
+fi
+
+--------------------------------------------------------------------------
```
MarkLodato: nit: split into a separate ```` ```yaml ```` codeblock.

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+# deploy_website.yml lives in .github/workflows.
+# This script is preliminary and for illustrative purposes. It may live in
+# a repo other than slsa.
+
+# ... Prepares the environment for Github Workflow
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
MarkLodato:
```suggestion
      - uses: actions/checkout@v3
        with:
          submodules: recursive
```

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+  bundle exec jekyll serve --livereload --incremental
+else
+  echo "Finish setting up code and installing dependencies. Exit."
+fi
+
+--------------------------------------------------------------------------
+
+# deploy_website.yml lives in .github/workflows.
+# This script is preliminary and for illustrative purposes. It may live in
+# a repo other than slsa.
```
MarkLodato: nit: I think the script has to live in the slsa repo

```suggestion
# This script is preliminary and for illustrative purposes.
```

0006/README.md
MarkLodato: note: if we're triggering on something other than the main branch, I think we'll need to add

```yaml
with:
  ref: main
```

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+# ... Prepares the environment for Github Workflow
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run the setup script
```
MarkLodato: For the rest of this, can we just use the https://github.com/actions/jekyll-build-pages? I think the only special thing we need to do to get our setup working is the `submodules: recursive` bit, right?

MarkLodato (commented):
Thanks! This plan LGTM. I just have some minor suggestions on the build setup, which doesn't significantly affect the proposal.

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+-   `/docs/spec`, `/docs/provenance`, `/docs/verification_summary`, and
+    `/docs/github-actions-workflow` are removed from the main branch.
+-   `/docs/images` doesn't contain versioned files and directories anymore. They
+    are removed or moved to their branch.
+-   `/docs/v1.0` and `/docs/v0.1` are two submodules that point to the
+    [version branches](#version-branches). As a result, `.gitmodules` is added
+    in the top level directory, and has `url = ./` for each submodule config to
+    ensure referencing to the local repo. A future version should follow a
+    similar setup in the repo.
```
MarkLodato: Can you add sections talking about the following?

- Process for users to contribute changes. They previously forked the repo, checked out `main`, made edits, then sent a PR to `main`. Now they will fork, check out `main` with submodules, make edits to the submodules(?), run some command to commit to a branch(?), then send a PR to the branch?

- Process for updating the submodule pointers. Will a bot do this automatically after every PR? How will that work?

MarkLodato (changes requested):
Oops, forgot something major!

0006/README.md
trishankatdatadog: Please let me know when this is done, and happy to review, thanks 🙂

0006/README.md
```diff
@@ -0,0 +1,319 @@
...
+-   All permalinks in file content use the dir structure described in
+    [the website](#the-website) section.
+
+NOTE: We admit the way of using the Git submodule is non-standard, but it is a
```
mlieberman85: Can you include something like a submodules best practices guide from somewhere. I know a lot of folks like myself have used submodules in the past and have run into a lot of issues.

0006/README.md
mlieberman85: Won't this lead to issues where something like main pulls down v2.0, v2.0 sees v1.0 and v0.1 as submodules. main see v1.0 and v0.1 as it's own submodules as well in its reference point and pulls that down? Especially if we're just looking at using git submodules against branches in the same repo and not splitting the repo up?

Wouldn't we need to create something like a `new_version_template` branch that didn't have the additional submodule logic in there otherwise we potentially end up with circular dependencies?

0006/README.md
chtiangg: I removed it and update the first bulletin for changing the way of clone the repo.

0006/README.md
chtiangg: Removed it.

0006/README.md
chtiangg: That is correct. I updated.

0006/README.md
chtiangg: Maybe, the example in "The Repo" section is not clear. I tried to use the example to show that:

"A version branch must only contain versioned files at its top level folder but no submodules, while the main branch must have the submodule folders in /docs but no versioned files."

I think it is an agreement here, so that the repo maintainer would not allow any submodule exists outside the main branch. Would you like me to explictly talk about the above statement in the "The Repo" section?

0006/README.md
chtiangg: Done.

0006/README.md
chtiangg: Ack. I was thinking just to use git switch in the middle of two build... but I feel we don't need to update this section since there is no example provided.

0006/README.md
chtiangg: Updated with two subsections: Code Contribution and Submodule Pointer. Could you please take a look? Thanks!

0006/README.md
chtiangg: I am sorry, I cannot find any docs that looks authorized to me from Google or Internal Google. It may because I am new to Git/GitHub and the open source world. Would you have a recommendation here?

In general, I feel it is best not to use submodule to update code. I added another two new sub-sections Code Contribution and Submodule Pointer. I hope they could be helpful to clarify my thoughts on my mind right now.

0006/README.md
MarkLodato: No, there is no recursion. The release branches do not have submodules, only the main branch. See the demo: https://github.com/chtiangg/slsa/tree/test/main

0006/README.md
```diff
@@ -0,0 +1,318 @@
...
+nice compromise and helps us to improve the maintenance experience.
+
+#### Version Branches
+
+Git branch is the way to manage versioned specifications.
+
+*What is changing*
```
mlieberman85, line 139: What process would need to be followed when creating a new version branch?

0006/README.md
```diff
@@ -222,69 +247,43 @@ Pages.
...
-    environment.
+    is updated to be:
+
+    ```bash
+    git clone --recurse-submodules https://github.com/chtiangg/slsa.git
+    git submodule update --remote
+    cd slsa/docs
+    ```
+-   A new Github Workflow script builds and deploys the website. The following
+    is a cod snapshot.
```
MarkLodato: nit: what is a "cod snapshot"?

0006/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Checkout repo, load all submodules, and build once
+
+The GitHub Workflow clones the `main` branch once, pulls the latest files for
+all the submodules, builds the website once, and deploys the website to Github
+Pages.
+
+*Disadvantages*
+
+-   The usage of Git submodule is non-standard, and it leaves some versioned
```
mlieberman85: What sorts of precautions do we need to take to make sure when using submodules to ensure we don't end up with cyclic dependencies?

Correct me if I'm wrong, but let's say for example we create SLSA v2-rc while still updating SLSA v1.5. If we have submodules pointing to those branches and if those branches themselves still contain the submodules you could end up with a situation where v2-rc points to v1.5 as a submodule and v1.5 points back to v2-rc right?

I would just suggest we enforce a fixed depth. Also if possible we avoid the branches themselves having references submodules and ensure only main has the references.

0006/README.md
```diff
@@ -0,0 +1,318 @@
...
+        - name: Build
+            uses: actions/jekyll-build-pages@v1
+            with:
+                source: ./docs
+        - name: Upload artifacts
+            uses: actions/upload-pages-artifact@v1
+    # ... Deploy to your own website.
+    ```
+
+### Alternatives considered: Build each branch separately, then assemble
```
mlieberman85: Is there a reason why vendoring of the versioned spec wasn't considered?

0006/README.md
```diff
@@ -206,6 +206,31 @@ Git branch is the way to manage versioned specifications.
...
+However, if the change comes from one of the repo owners, and it has to modify
+files in both `main` and version branches in one PR, we assume the owner
+understands the risk and knows how to send a proper code update with Submodule
+changes.
+
+#### Submodule Pointer
+
+A new change in a version branch makes the Submodule pointer in the `main`
+branch stale. A Github Workflow script may be created to automatically update
+the Submodule pointing to the latest commit, when the PR is merged. If desired,
```
MarkLodato: Can we make this part of the plan, to have a workflow that automatically submits a change to update the submodule? I don't think anyone will have the rigor to do it manually.

0006/README.md
```diff
@@ -206,6 +206,31 @@ Git branch is the way to manage versioned specifications.
...
+Since the repo uses the Git submodule in a non-standard way and for easy
+building website purposes, people SHOULD NOT use submodules to commit any
+change.
+
+When a community member needs to make a contribution, they fork the repo (and
+clone it locally if they are familiar with git). If the change is for a
+versioned file (a.k.a. spec file), they should make the change in the version
+branch, and send a PR to the version branch. Meanwhile, if the change is for an
+unversioned file or the website itself, they should make the change in `main`,
+and send a PR to `main`.
```
MarkLodato: I don't think this is workable as-is. If you check out a release branch, you can't preview the changes with Jekyll since it just contains Markdown but not the full site. To preview you need to check out `main` and update the submodule. But if you edit there, it's really hard to commit it as a branch.

Upon second thought, maybe submodules aren't a great fit here. Let me try something out...

0006/README.md
chtiangg: Would it be an agreement between repo owners, when a new version branch is needed? I assume each prior version was created when there was a discussion during the SLSA meeting. I personally feel it is a step before this doc which focuses on improving maintenance experience to cleanup the repo structure and automate the build, while defining the whole process of what everyone needs to do seems beyond the scope. What do you think?

0006/README.md
chtiangg: I am sorry, I don't understand the meaning of "vendoring of the versioned spec". 

Do you mean using different repos for different versions (instead of branches)? In my opinion, a repo for 10~ markdown pages is an overkill, but you may mean something else.

MarkLodato:
Here's an alternative to git submodules, using `git worktree` instead:

```bash
# Same as before:
$ git clone https://github.com/slsa-framework/slsa
$ cd slsa
# And then (could be done with a script):
$ for version in v0.1 v1.0; do git worktree add docs/$version release/$version; done
```

(Also add `/docs/v1.0` and `/docs/v0.1` to .gitignore, which is a one-time thing.)

Now you edit the files in `docs/v1.0` like normal, except that the directory _has its own checkout_. Doing `git status` in the parent directory doesn't show any changes in the version directories, which is unfortunate.

Benefits:

- Over submodules:
  - No need to keep the `main` branch up to date after each version branch commit.
  - You can preview changes as you edit.
  - Git-foo is not quite as difficult. For example, there is no detached head, which confuses everyone.
- Over status quo:
  - We can track changes to versions with branches.

Drawbacks;

- Over both submodules and status quo:
  - Building from `main` is no longer deterministic since the specific version is not encoded.
  - `git status` doesn't show changes to the submodule
- Over status quo:
  - Still much harder to edit.

I'm still not sold, but maybe put it in alternatives considered?

MarkLodato:
To be clear, there are three (mostly) independent parts of this proposal:

- Changing the URL structure, e.g. slsa.dev/spec/v1.0 becomes slsa.dev/v1.0.
- Switching to GitHub Actions to build.
- Tracking versions with git branches.

Here's a possible solution to just the branch question using `git subtree`:

- Keep `main` as-is, with each version as a separate directory. People edit as they do today.
- Create `release/<version>` using `git subtree`, and keep it up-to-date automatically using GitHub Actions.

Effectively each `release/<version>` branch is the history just of that one directory `main:/docs/<version>/`. Uses:

- Tracking changes across versions by looking at the git log. (Also has the benefit of filtering out changes that don't touch the content.)
- Creating new versions. Instead of doing a directory copy, we create a new branch and add a new subtree.
- Merging changes between versions, e.g. during a release candidate.

I'm now leaning toward this...

*Edit:* If we go this route, we may want to hand-craft the branch history since it's been lost due to the directory copies. I'd be happy to help with that.

0006/README.md
chtiangg: I agree and think maybe people operate very differently from I do. I tend to deal with one small thing at a time. When I think about making spec changes, regardless of big or small, that is really a markdown and it has already had a markdown preview in github.com. Including this spec change in the website + preview the website is another change, and it feel like a responsibility from the repo owner.

(And yes, thanks Mark for helping with other alternatives!)

0006/README.md
chtiangg: I will wait and cleanup this section after our discussion :)

0006/README.md
chtiangg: Will also update this after our meeting.

0006/README.md
chtiangg: Let me copy part of the example above from the "The Repo" section.

```
slsa-framework/slsa:main
├─ docs/
│  ├─ ...
│  ├─ v0.1/    # a submodule points to the branch `release/v0.1`.
│  ├─ v1.0/    # a submodule points to the branch `release/v1.0`.
│  ├─ ...
│  └─ <other-unversioned-files>.md
├─ README.md
└─ ...

slsa-framework/slsa:release/v1.0 ===> ASSUME THIS IS  `SLSA v2-rc` from YOUR EXAMPLE
├─ ...
└─ <v1.0-versioned-files>.md

slsa-framework/slsa:release/v0.1 ===> ASSUME THIS IS  `SLSA v1.5` from YOUR EXAMPLE
├─ ...
└─ <v0.1-versioned-files>.md
```

Every version branch has a submodule that exists exactly once in main. Updating `SLSA v2-rc` and `SLSA v1.5` has nothing to do with submodules. Both version branches are self-contained and should only have spec files, and maybe some images (If a PR tries to violate the rule, maybe the repo owner should reject it?). The updated "Code Contribution" section explicitly mentioned that submodules should not be used to commit any changes. With that said, the depth is going to be one, and no submodule exist outside the main branch. However, please let me know if I missed any corner case here.

Mark and I will meet again to discuss the other subtree solution. You may see an update coming soon.

0006/README.md
chtiangg: Done.

0006/README.md
chtiangg: Didn't see your response, and assume you are OK with it.

Also, I updated the doc to propose using git subtrees instead. I will later resolve this comment to reduce noise.

0006/README.md
chtiangg: Resolving this conversation to reduce noise.

0006/README.md
chtiangg: Updated the content to propose git subtree instead. I will resolve this comment.

0006/README.md
chtiangg: Updated the content to propose using Git subtrees.

0006/README.md
chtiangg: Resolving this comment to reduce noise.

0006/README.md
chtiangg: Added an another variant for using a repo to manage a specific version. I will resolve this comment.

0006/README.md
chtiangg: I added a bulletin to talk about sync a git subtree change to its branch.

mlieberman85:
So, I'm reading through this, and I'm still not really understanding why it has to be so complicated.

I'm on board with changing url structure and moving to github actions for more complicated site builds.

Using git subtrees or submodules that point to the same repo, just different branches still isn't clicking for me yet. I do find subtrees a bit easier to manage than submodules. Do we have an actual workable example showing this. I'm worried that it will lead to confusion around cyclic dependencies. It would also complicate how new branches are cut.

For example, from the doc it is really unclear how when doing this we are intended to prevent:

1. `main:docs/v1` points to branch `release/v1:docs/v1`
2. We cut a branch from main for `release/v2`
3. `release/v2` would have `release/v1` as a submodule which is unintended.

From the proposal it's not clear how this is intended to be solved? Is the idea here to have `release_template` branch? Is the idea here now that main is only used purely for the structure of updating the site? All non-released changes should still go to some draft branch?

There's a lot of unanswered questions.

0006/README.md
```diff
@@ -0,0 +1,364 @@
...
+    -   However, using `baseurl: "/<version>" destination: "_site/<version>"` in
+        `_config.yml` can avoid moving build results around.
+
+*Variant 1*
+
+Clone the repo in different jobs, build a specific branch in each job, assemble
+the website, and deploy it. This option requires sharing build results between
+jobs in Github Workflow, so it is more complicated.
+
+*Variant 2*
```
mlieberman85, line 316: So, when I said a separate repo. I did not mean a separate repo for every version. It doesn't even have to be a separate repo. We can just vendor the files from the other versions as releases and as part of the build we can just pull those files into the github actions build.

Separately if we do go submodules I really recommend us using a different repo with different branches or whatever to prevent cyclic dependencies and inadvertently pulling in submodules in places they were not intended, e.g. v3 pulls in the submodule for v2 or whatever.

chtiangg: > Do we have an actual workable example showing this. I'm worried that it will lead to confusion around cyclic dependencies. It would also complicate how new branches are cut.

Yes, for Git submodule, but No, for Git subtree. When I received the feedback from Joshua, we've tested Git submodules and mentioned it in [google doc](https://docs.google.com/document/d/1sTo6Lccu5I5mEY8oemF964KbKQvdvUHcbVRdcCxK24w/edit#heading=h.17wg41voij6q) which is listed at the very top of this proposal. I copy and pasted the section from the doc below:

As a proof-of-concept, in Tianyu Chen’s Github repo, with similar logic, they used this [Github Workflow](https://github.com/chtiangg/chtiangg.github.io/blob/main/.github/workflows/deploy_the_slsa_website.yml) which runs [this script](https://github.com/chtiangg/chtiangg.github.io/blob/main/.github/scripts/build.sh) to deploy the SLSA website to [his personal website](https://chtiangg.github.io/).

> I'm worried that it will lead to confusion around cyclic dependencies. It would also complicate how new branches are cut.

With Git subtree, I am not sure where the cyclic dependencies come from, while the complicated parts are all workload to the repo owners (in my mind, they are Mark and Joshua). I'd assume they two are the one to create new branches, and add Git subtrees, while it is agnostic for other people when they are looking at main. Right now, this doc is proposing Git subtree rather than Git submodule.

> From the proposal it's not clear how this is intended to be solved?

Would you have any suggestion for the "Example" under the "The Repo" section? I am happy to make the example more clear to show the desired status for the repo.

> Is the idea here to have release_template branch? Is the idea here now that main is only used purely for the structure of updating the site? All non-released changes should still go to some draft branch?

No, No, and No. The "Code Contribution" section talks that "The process should remain the same as it is for today.".  Maybe I should delete the wording "as it is for today" in the sentence? What do you think?

To answer having release_template branch, or what is going into a version branch, I think it again falls into the responsibility of repo owner (Mark and Joshua). I'd trust both understand how to setup the repo to make this statement true: "The process should remain the same as it is for today.".

> We can just vendor the files from the other versions as releases and as part of the build we can just pull those files into the github actions build.

It honestly sounds like a lot of effort to assemble the input of building the website, which is one of the primary reasons we try to avoid by using subtree and submodule.

lehors:
I must admit not to have experience with submodules and subtrees so I will leave it to you guys to decide which might be best. I'm happy to be the guinea pig and try things out though.
That being said, as Mark pointed out, there are three different aspects being discussed and I wonder whether we shouldn't try to tackle these one at a time instead of all at once. I realize this would imply some extra work because implementing the next change will require making some changes to what was done for the previous one but it still seems desirable to me.

MarkLodato:
> That being said, as Mark pointed out, there are three different aspects being discussed and I wonder whether we shouldn't try to tackle these one at a time instead of all at once.

Yeah, that sounds good. In fact, I think only the branch strategy change needs a design doc.

- The build process change can just be a PR, since it's a very minor change that most people won't see.
- The URL change can be discussed on a GitHub issue. I don't expect a lot of back-and-forth here.

> I'm still not really understanding why it has to be so complicated.

It seems like it should be easy, right? But when you dig into it, there's no obvious solution, at least as far as I can tell.

This whole project came from the premise, "Shouldn't we be using branches instead of directory copies?" In other words, each version of the spec lives in a separate git branch, and then we somehow assemble the different versions into a single site. We could either build each branch separately then assemble the built artifacts, or assemble the sources and build once. Both of these are more difficult for contributors than what we have today.

So that's where `git subtree` comes in. In this model, we keep developing the spec **exactly as we have been**, with everything going in `main`. The per-version branches only exist to either (a) view the version history or (b) do maintainer stuff like creating a new version or merging between versions. Most people don't have to know it exists.

More specifically:

- `main` looks exactly as it does today.
- `release/<version>` contains only the files in `/docs/spec/<version>`

This can be accomplished with `git subtree split --prefix=docs/spec/v1.0 -b release/v1.0`, though in practice we might want to manually clean up the history.

trishankatdatadog:
I am OK with the changes are Mark has [summarised](https://github.com/slsa-framework/slsa-proposals/pull/10#issuecomment-1557482563).

chtiangg:
Thanks everyone for commenting the feedback here. To share, I have a concrete example of using git subtree in my fork:

- Repo: https://github.com/chtiangg/slsa
  - branch: subtree/main (mimic what is going into main), subtree/v1.0, and subtree/v0.1
- Deployed website (on my personal website):  https://chtiangg.github.io/
- Git Workflow script: https://github.com/chtiangg/chtiangg.github.io/blob/main/.github/workflows/build_subtree_sites.yml

The three subtree branches match the example in the "The Repo" section. I also started to experiment with syncing a change in subtree/main to subtree/v1.0 and the reverse way.

> Yeah, that sounds good. In fact, I think only the branch strategy change needs a design doc.
> - The build process change can just be a PR, since it's a very minor change that most people won't see.
> - The URL change can be discussed on a GitHub issue. I don't expect a lot of back-and-forth here.

I think it also depends on if the proposal should tell a full story. 

In my opinion, moving sections other than branches to a "reference" section adds a lot of clarity. Otherwise, after a couple of years, I don't think people can tell what happened by just looking at a branch change proposal. Even if we list bugs and other discussion in the doc, I'd also imagine no one would be willing to dig into to understand the connection or rationales.

However, Mark, could you please suggest here, since reviewers start to align with your comments? Thanks!

MarkLodato:
> In my opinion, moving sections other than branches to a "reference" section adds a lot of clarity. Otherwise, after a couple of years, I don't think people can tell what happened by just looking at a branch change proposal. Even if we list bugs and other discussion in the doc, I'd also imagine no one would be willing to dig into to understand the connection or rationales.

@mlieberman85, @lehors, @trishankatdatadog Any thoughts?

trishankatdatadog:
> > In my opinion, moving sections other than branches to a "reference" section adds a lot of clarity. Otherwise, after a couple of years, I don't think people can tell what happened by just looking at a branch change proposal. Even if we list bugs and other discussion in the doc, I'd also imagine no one would be willing to dig into to understand the connection or rationales.
> 
> @mlieberman85, @lehors, @trishankatdatadog Any thoughts?

Sorry, I don't understand this particular proposal. @chtiangg, would you please elaborate with an example?

chtiangg:
> > > In my opinion, moving sections other than branches to a "reference" section adds a lot of clarity. Otherwise, after a couple of years, I don't think people can tell what happened by just looking at a branch change proposal. Even if we list bugs and other discussion in the doc, I'd also imagine no one would be willing to dig into to understand the connection or rationales.
> > 
> > 
> > @mlieberman85, @lehors, @trishankatdatadog Any thoughts?
> 
> Sorry, I don't understand this particular proposal. @chtiangg, would you please elaborate with an example?

I think in general, a proposal gives people the knowledge to make better decisions. It helps us influence and predict the success of the design, and navigate risk and find solutions. Telling a full story is a way to achieve so. Once a proposal is agreed and understood by stakeholders, it can be then broken down into small tasks and issues - bugs, issues, sub-tasks, discussion, and code changes . While the detailed implementation may be different from a proposal, since no one could cover everything in a doc without really working out the concrete thing, a proposal is still useful to give overall guidance and let people understand "The overall rationales triggered people to make the decision at that particular time".

Even if this doc should mainly talk above the repo change, website and URL change, and Build options are all part of the full story. So, I am suggesting highlighting the repo change part (maybe put the change in the first few sections), while adding an extra section called "Reference", move the website change and Github Action build script change into it. Alternatively, just move "The Repo" section in front of others to keep a full story in one single doc. An example TOC of this doc could be:

```
- Abstract
- Problem Statement
- Design Idea
  - The Repo
- Reference
  - The Website 
  - The Github Workflow Script 
- Alternatives considered
```

Please let me know if my explanation makes sense to you. Thank you!

trishankatdatadog:
> Please let me know if my explanation makes sense to you. Thank you!

Yes, sounds good to me, thanks

0006/README.md
```diff
@@ -0,0 +1,364 @@
...
+The process should remain the same as it is for today. Github subtrees and
+version branches are agnostic to community contributors.
+
+-   If a new version is desired, a contributor should raise the request either
+    in the SLSA weekly meeting or create an issue in the repo. The repo owner
+    evaluates the request, and sets up the new version in the repo.
+-   If a change is desired in the prior or future version, the contributor
+    should fork the repo, clone it locally if they are familiar with git, make
+    the change in the `main:docs/<version>` directories, and send a PR against
+    `main`.
```
lehors, line 218: I don't understand this. If I want to make a change to v0.1 for instance. Wouldn't I need to submit my change against the release/0.1 branch rather than main?

lehors:
> I think in general, a proposal gives people the knowledge to make better decisions. It helps us influence and predict the success of the design, and navigate risk and find solutions. Telling a full story is a way to achieve so. 

Ok but you choose to see this as one story, you could equally have three different stories:

- Changing the URL structure, e.g. slsa.dev/spec/v1.0 becomes slsa.dev/v1.0.
- Switching to GitHub Actions to build.
- Tracking versions with git branches.

:-)

But that's ok. I don't think we need to get hang up on that and spend more time on how the document ought to be structured. What matters is to get a plan that we think will work and are comfortable with.
I think I now understand what the git subtree approach entails and I'm good with it.
Thanks.

chtiangg (commented):
Since I didn't receive any new feedback during the past few days, I made the following changes according to the latest comment/agreement.

Kept the section for 
- Tracking versions with git branches.

Moved these two sections into "Appendix":
- Changing the URL structure, e.g. slsa.dev/spec/v1.0 becomes slsa.dev/v1.0.
- Switching to GitHub Actions to build.

Please take a look, Thank you!

trishankatdatadog:
> Please take a look, Thank you!

LGTM. One more thing: so the template for the pages across all the MAJOR.MINOR versions will be the same, correct (since they are not versioned, so to speak, by git subtrees)? So if the template changes, the rebuilt pages will look the same across all versions?

chtiangg:
Thanks again for reviewing this proposal!

> LGTM. One more thing: so the template for the pages across all the MAJOR.MINOR versions will be the same, correct (since they are not versioned, so to speak, by git subtrees)?

Yes, you are correct. It is also my current understanding of the slsa website, such as template code, css, configs and so on.

>  So if the template changes, the rebuilt pages will look the same across all versions?

Yes, the look-and-feel stays the same. I should add that the content in the navigation panel (or table of content) may be different, and it is true even as of today, because two versions (for the specification) have different pages, and different navigation configs that are NOT living in their versioned directories (e.g. v1.0 or v0.1) or in any future subtree branches.

lehors:
The style of a page is governed by the "layout" used, defined in the file header as in "layout: specifications". We have several layouts defined in docs/_layouts. So this is indeed outside the versioned directories but different versions could use different layouts if we want. I don't think that's a problem.

lehors (approved):

joshuagl:
I really like the switch to `git subtree`, that sub-command is new to me but looks like exactly what we need.

Will we want/need to adopt some kind of commit convention to make it easy to extract a released spec from the main branch to a new subtree?

MarkLodato:
> Will we want/need to adopt some kind of commit convention to make it easy to extract a released spec from the main branch to a new subtree?

I don't anticipating needing to, though I still would like to do so independently as part of https://github.com/slsa-framework/slsa/issues/821.

0006/README.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+### Main Branch
+
+The directory structure of `main` stays the same as the repo has right now.
+
+*What is changing*
+
+-   `/docs/spec`, `/docs/provenance`, `/docs/verification_summary`, and
+    `/docs/github-actions-workflow` are removed from `main`.
```
MarkLodato: No need to touch `github-actions-workflow`. It has moved out of this repo and those files are just redirects.

Same below.

0006/README.md
```diff
@@ -0,0 +1,373 @@
...
+The process should remain the same. Syncing Github subtrees and version branches
+are agnostic to community contributors.
+
+-   If a new version is desired, a contributor should raise the request either
+    in the SLSA weekly meeting or create an issue. The repo owner evaluates the
+    request, and sets up the new version in the repo.
+-   If a change is desired in the prior or future version, the contributor
+    should fork the repo, clone it locally if they are familiar with git, make
+    the change in the `main:docs/<version>` directories, and send a PR against
+    `main`.
```
MarkLodato: I suggest splitting this out between regular contributions and creating a new version, and also writing out the commands to create the new branch.

`````suggestion
The process should remain the same.  Contributors fork, clone, and submit PRs
to the version directories in `main` just as they do today. Syncing Github
subtrees and version branches should be transparent to community contributors.

### Creating new versions

When a new version is desired, a project maintainer can create a new branch as
follows (using `v1.0` → `v1.1` as an example):

```bash
git branch release/v1.1 release/v1.0
git checkout main
git subtree add --prefix=docs/spec/v1.1 release/v1.1
git push origin release/v1.1 main
```

The GitHub Actions workflow that syncs branches should automatically pick up the new version.
`````

File: 0006/README.md (hunk 14 above)
Comment (chtiangg) [reply 0006/README.md:316]: Assume you are OK here since I didn't hear anything for a week.

File: 0006/README.md [hunk 18]
```diff
@@ -0,0 +1,373 @@
...
+      - name: Build
+          uses: actions/jekyll-build-pages@v1
+          with:
+              source: ./docs
+      - name: Upload artifacts
+          uses: actions/upload-pages-artifact@v1
+# ... Deploy to your own website.
+```
+
+### Disadvantages of The Proposed Mechanism
```
MarkLodato: I think everything from here on down should be part of the main doc, not the appendix.

0006/README.md
```diff
@@ -0,0 +1,373 @@
...
+    in the SLSA weekly meeting or create an issue. The repo owner evaluates the
+    request, and sets up the new version in the repo.
+-   If a change is desired in the prior or future version, the contributor
+    should fork the repo, clone it locally if they are familiar with git, make
+    the change in the `main:docs/<version>` directories, and send a PR against
+    `main`.
+
+## Appendix
+
+### The Website
```
MarkLodato: Can you add a short paragraph explaining that these changes are independent of this proposal here, but that we will want to do them first because they affect how we create branches? And also link to the relevant issues for those?

0006/README.md
```diff
@@ -0,0 +1,373 @@
...
+
+Out of scope:
+
+-   [out of scope] Change the static site generator used to build the website,
+    such as upgrading to Jekyll v4~ or switching to Hugo, though some work here
+    unblocks that.
+-   [out of scope] Make major updates to the content of any files in the repo,
+    although fixing URLs with the proposed website structure is in scope.
+
+## The Repo Structure of slsa-framework/slsa
```
MarkLodato: Could you add a section talking about what the benefits of this proposal are, and how we intend to use the new branches? In other words, why are we bothering with this? Basically (please verify all of this is true):

(1) To allow viewers to more easily see the revision history of the specification across SLSA versions. 

Currently it is difficult to view the history of changes to the specification across versions due to the directory copies. If you limit the git history to the `/docs/spec/<version>` directory, this history stops at the copy ([example](https://github.com/slsa-framework/slsa/commits/main?after=750dc05dc88c81aa9afc35beed1faaa9ed6aabde+0&branch=main&path%5B%5D=docs&path%5B%5D=spec&path%5B%5D=v1.0&qualified_name=refs%2Fheads%2Fmain)). If you look at the entire repo, it is hard to follow because there are changes to many parts of the site all mixed together. This makes it hard to actually follow how SLSA changed over time.

With this proposal, one can view the history by simply viewing the git log of the `release/<version>` branch. This branch only contains commits that went into `<version>`, and it includes history from prior branches. It also allows ones to easily diff between two versions since they have a common ancestor.

(2) To make it easier to develop and release new versions.

Currently it is somewhat painful to develop a new version due to the directory copies. Changes from one branch cannot easily be merged into another branch, making it difficult to keep an upcoming version in sync with changes to the latest stable version. When it comes time to cut a release candidate, we have to do directory copies and hand-edit a bunch of files, and it's hard for reviewers to ensure that nothing has been changed accidentally.

With this proposal, development of a new version is more natural. Creating a branch is done via `git branch` (plus `git subtree add`), merging between branches is possible (with `git subtree merge`), and cutting a release only involves creating a new subtree.

MarkLodato (changes requested):

0006/README.md
chtiangg: Thank you Mark!

I added a section "Benefits of This Proposal" and used the content from this comments. I agree and believe both bulletins are true (although I haven't played with `git subtree merge` before).

0006/README.md
chtiangg: Removed for both.

0006/README.md
chtiangg: Added the paragraph right below the title "Appendix".

However, I don't know the issue for using the GIthub Actions workflow to build website, do you happen to know one or shall I create an issue for it? Thanks!

0006/README.md
chtiangg: Moved.

chtiangg (commented):
Updated the doc, although I haven't performed the squash operation for old commits yet.

trishankatdatadog (approved):
LGTM

trishankatdatadog (approved):
LGTM

trishankatdatadog (approved):
LGTM

trishankatdatadog:
P.S. Jesus, not sure what happened there, only meant to approve once, so I blame a race condition somewhere...

joshuagl:
> > Will we want/need to adopt some kind of commit convention to make it easy to extract a released spec from the main branch to a new subtree?
> 
> I don't anticipating needing to, though I still would like to do so independently as part of [slsa-framework/slsa#821](https://github.com/slsa-framework/slsa/issues/821).

I had misunderstood an aspect of the proposal. I was originally imagining that the next version of a spec is developed in `main` and split into a subtree at release time, however I think what's actually being proposed is that all version of the spec, draft or approved, are developed in branches. That would certainly make my idea of conventional commits to identify spec changes less relevant to this proposal.

chtiangg:
> I was originally imagining that the next version of a spec is developed in main and split into a subtree at release time, however I think what's actually being proposed is that all version of the spec, draft or approved, are developed in branches.

I think it should be in both ways. 

```
git branch release/v1.1 release/v1.0
git checkout main
git subtree add --prefix=docs/v1.1 release/v1.1
git push origin release/v1.1 main
```

The 3rd steps above add a sub-fold `docs/v1.1` into `main`.  We and any community contributor could still work in this `docs/v1.1` and make changes. A Github Actions worflow syncs the change to the branch.

Meanwhile, working in the `release/v1.1` can be synced back to `main` into `docs/v1.1`  directory as well. However, the workflow above is definitely in higher priority, because it ensures the current contribution/workflow doesn't change.

joshuagl (approved):

chtiangg:
@joshuagl 

Just a F.Y.I., I added concrete examples for the git subtree command to sync between a subtree branch and a subtree directory.

```bash
# If a change is committed into `main`, the workflow runs
git subtree push --prefix docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1

# if a change is committed into `release/v1.1`, the workflow runs
git subtree pull --prefix docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1 --squash
```

joshuagl:
> Just a F.Y.I., I added concrete examples for the git subtree command to sync between a subtree branch and a subtree directory.

awesome, thanks for that.

0006/README.md
```diff
@@ -0,0 +1,432 @@
...
+    only needs to make changes in `main`.
+-   When detecting a push in the versioned directories, a Github Actions
+    workflow syncs the same change into the version branches. For example:
+
+    ```yaml
+    # sync_branch.yml lives in .github/workflows.
+    on:
+      push:
+        branches: [ main ]
+        paths: [ docs/v+/ ]
```
MarkLodato:
```suggestion
        paths: [ "docs/v[0-9.]+/**" ]
```

0006/README.md
```diff
@@ -0,0 +1,432 @@
...
+
+The GitHub Actions workflow that syncs branches should automatically pick up the
+new version, for example:
+
+```bash
+# If a change is committed into `main`, the workflow runs
+git subtree push --prefix docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1
+
+# if a change is committed into `release/v1.1`, the workflow runs
+git subtree pull --prefix docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1 --squash
```
MarkLodato: nit: for consistency with the command above

```suggestion
# If a change is committed into `main`, the workflow runs
git subtree push --prefix=docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1

# if a change is committed into `release/v1.1`, the workflow runs
git subtree pull --prefix=docs/v1.1 https://github.com/slsa-framework/slsa.git release/v1.1 --squash
```

MarkLodato: Alternatively, we could make this a one-directional sync, and disallow any commits to `release/<version>`.

MarkLodato (approved):
Let me know when it's ready to merge.

0006/README.md
chtiangg: Done

mlieberman85 (approved):
LGTM

PR #15, opened 2023-07-21
Add proposal Steering Committee application, terms, and role

joshuagl:
This proposal clarifies the SLSA Steering Committee, specifically with regards to:

* responsibilities
* difference to core maintainer responsibilities
* application, nomination and terms

### DIFF
--- a/0007/README.md
+++ b/0007/README.md
@@ -131,11 +131,13 @@ duties for each role.
     *   Some Core Maintainers might not be SC Members.
     *   The SC, as a group, are responsible for maintaining the relationship
         with the SCI WG, which includes:
-        *   Facilitating input from SCI WG into SLSA during significant 
-            developments, including: adding tracks and establishing roadmaps
-        *   Periodically reporting progress and developments to SCI WG 
+        *   Facilitating input from SCI WG into SLSA during significant
+            developments, including: adding tracks and establishing roadmaps.
+        *   Periodically reporting progress and developments to SCI WG.
+        *   Sharing relevant information from the SCI WG with the SLSA community.
     *   As community leaders, SC Members are also responsible for maintaining
         the slsa-framework/governance repository.
+    *	SC Members share updates on their activities in the Community Meeting.
 
 ### Clarify GitHub roles and how they maps to project roles
 
--- a/0007/README.md
+++ b/0007/README.md
@@ -17,7 +17,7 @@ contributors.
 
 ## Problem statement
 
-The original SLSA Steering Committee has passed its two year term and, per
+The original SLSA Steering Committee (SC) has passed its two year term and, per
 [current governance documents][current terms], must now select the next group
 of Steering Committee members. This gives us a natural point for reflection on
 the role and operation of the Steering Committee. Several project contributors
@@ -76,6 +76,8 @@ duties for each role.
     submitting a brief (1-2 paragraph) application consisting of:
     *   **Evidence** of performing some of the role's duties over the recent
         past.
+        *   For SC Member applications, this should include evidence of being
+            well-versed in SLSA.
     *   **Statement of intent** to continue performing duties over the next
         year.
     *   **Affiliation** / employer.
@@ -123,11 +125,10 @@ duties for each role.
     project strategy, and serve as an escalation point in the event of a
     disagreement.
     *   These duties are much less frequent than those of a Maintainer.
-    *   We generally **expect an SC Member to also be a Core Maintainer**, or a
-        Subproject Maintainer for a critical subproject, in order to ensure
-        that SC Members are well versed in the project.
-    *   Some Core Maintainers might not be SC Members. This may be a stepping
-        stone to becoming an SC Member.
+    *   We **expect an SC Member to be well versed in the project**. This often
+        means that SC Members are also Core Maintainers, or maintainers of a
+        subproject, but this is not required.
+    *   Some Core Maintainers might not be SC Members.
     *   The SC, as a group, are responsible for maintaining the relationship
         with the SCI WG, which includes:
         *   Facilitating input from SCI WG into SLSA during significant 
--- a/0007/README.md
+++ b/0007/README.md
@@ -198,18 +198,17 @@ GitHub organisation:
 *   Status quo
     *   see [goals]
 *   Dissolve the SLSA Steering Committee in favor of steering from the SCI WG
+    *   Pro: simplified governance structure.
     *   Con: Strategy may be driven by those without deep familiarity with
         the project.
     *   Con: SLSA community may not feel an appropriate sense of ownership and
         autonomy.
 *   Merge the Core Maintainer and Steering Committee Member roles
-    *   Pro: simplified structure
-        *   Counterpoint: the simplification does not result in too significant
-            a savings
+    *   Pro: simplified structure.
     *   Con: prevents us from having Core Maintainers who are not SC Members or
-        vice versa, which may be desirable
+        vice versa, which may be desirable.
     *   Con: technically the two roles are distinct, even if the same
-        individuals have typically performed them
+        individuals have typically performed them.
 
 <!-- References -->
 
--- a/0007/README.md
+++ b/0007/README.md
@@ -192,6 +192,8 @@ GitHub organisation:
 *   Add requirement to publish minutes for every SC meeting.
 *   State explicitly that SLSA lives within and takes broad direction from the
     OpenSSF SCI WG.
+*   Document a tie-breaking method for SC votes, should they be deadlocked, i.e
+    when the SC has an even number of members.
 
 ## Alternatives considered
 
--- a/0007/README.md
+++ b/0007/README.md
@@ -1,0 +1,220 @@
+# Proposal 7: Steering Commitee application, terms, and role
+
+*   Proposer: [Joshua Lock](https://github.com/joshuagl)
+    (joshuagloe@gmail.com on behalf of joshua.lock@uk.verizon.com),
+    [Mark Lodato](https://github.com/MarkLodato) (lodato@google.com)
+*   GitHub Issue: N/A
+*   Status: [DRAFT](../README.md#meaning-of-status-codes)
+*   Original link:
+    [Google docs](https://docs.google.com/document/d/1DBB-ziaM0yHyL1DDw1my-qs7tgKxluK25EIYTQltcgY/edit?usp=sharing)
+    (see that doc for reviewer comments)
+
+## Abstract
+
+This document proposes a change of SLSA governance structure to reduce
+redundancy, ensure transparency, and improve recognition for invested
+contributors.
+
+## Problem statement
+
+The original SLSA Steering Committee has passed its two year term and, per
+[current governance documents][current terms], must now select the next group
+of Steering Committee members. This gives us a natural point for reflection on
+the role and operation of the Steering Committee. Several project contributors
+have raised justifiable concerns about the lack of transparency of the
+committee, the ambiguity in how responsibilities differ for the role of a
+Steering Committee member vs. Maintainers, and what the relationship is with
+SLSA’s parent [Supply Chain Integrity Working Group (SCI WG)][SCI WG].
+
+### Historical context
+
+The SLSA Steering Committee was formed during the early days of both SLSA and
+the OSSF to provide a diverse set of inputs and oversight to the specification.
+Since the committee's original formation, the OpenSSF has developed more
+[structure][OSSF structure] and [processes][OSSF processes] which could
+support the SLSA project.
+
+## Prior art
+
+Many governance models exist for steering or umbrella projects, i.e.; the
+current [SLSA Steering Committee], [Kubernetes Steering Council],
+numerous [other models] within the CNCF.
+
+## Goals
+
+[goals]: #goals
+
+*   High-level vision/strategy for the SLSA project should be shaped through
+    diverse inputs, including from voices outside of the SLSA project’s active
+    contributors.
+*   Transparency at all levels of SLSA governance, with clearly defined roles
+    and responsibilities.
+*   Greater recognition of existing, and future, active contributors.
+*   Strong sense of ownership for active contributors.
+*   Diverse affiliation of community leadership, such that a single entity
+    cannot unduly influence the project.
+*   Lower cost of operations for SLSA governance. The SLSA organization is
+    small, so a heavyweight process is not worth the effort. In addition, SLSA
+    can lean on the [established organizational support][OSSF structure] in the
+    OpenSSF if needed.
+
+
+## Proposed solution
+
+Below is a summary of the proposed changes, highlighting the important
+differences in **bold**. The purpose of this proposal is to get general
+agreement on the idea; exact wording will be worked out in the implementation
+PR. See [current governance doc] for reference, including the list of example
+duties for each role.
+
+### Replace “terms” / “seats” with a lightweight re-application process
+
+*   SC **no longer has 7 fixed seats**, but instead has a minimum of 3 and
+    ideally about 5.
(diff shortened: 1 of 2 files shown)


0007/README.md
```diff
@@ -0,0 +1,220 @@
...
+*   SC Member: a **senior leader** of the SLSA project who is **trusted** and
+    **expected** to ensure the health of the project overall, drive the overall
+    project strategy, and serve as an escalation point in the event of a
+    disagreement.
+    *   These duties are much less frequent than those of a Maintainer.
+    *   We generally **expect an SC Member to also be a Core Maintainer**, or a
+        Subproject Maintainer for a critical subproject, in order to ensure
+        that SC Members are well versed in the project.
+    *   Some Core Maintainers might not be SC Members. This may be a stepping
+        stone to becoming an SC Member.
```
MarkLodato: Suggestion from @brunodom was to soften this language. Maybe just say that:
- We expect SC Members to be well versed in the project (or to show expertise, or something like that)
- In practice SC Members are likely also Core Maintainers, but this is not a requirement.
- Remove the sentence about being a stepping stone.

0007/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Abstract
+
+This document proposes a change of SLSA governance structure to reduce
+redundancy, ensure transparency, and improve recognition for invested
+contributors.
+
+## Problem statement
+
+The original SLSA Steering Committee has passed its two year term and, per
```
MarkLodato:
```suggestion
The original SLSA Steering Committee (SC) has passed its two year term and, per
```

So make it clear what "SC" means in this doc.

0007/README.md
joshuagl: [aff2902](https://github.com/slsa-framework/slsa-proposals/pull/15/commits/aff29026d276b7dbb8659f75224ecf80cc3203ae) softens the language, as suggested.

@brunodom @MarkLodato PTAL.

kimsterv:
+1 Lgtm!

0007/README.md
brunodom: LGTM.

MarkLodato (approved):
@brunodom and @kimsterv could you approve via the GitHub interface (Files Changed -> Review changes -> Approve) so that others can easily see the checkmark in the upper right? Thanks!

kimsterv (approved):

trishankatdatadog (approved):

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+agreement on the idea; exact wording will be worked out in the implementation
+PR. See [current governance doc] for reference, including the list of example
+duties for each role.
+
+### Replace “terms” / “seats” with a lightweight re-application process
+
+*   SC **no longer has 7 fixed seats**, but instead has a minimum of 3 and
+    ideally about 5.
+    *   Same guidelines for Maintainers, without the hard minimum.
+*   Individuals **self-nominate** to become a Maintainer or SC Member by
```
melba-lopez, line 75: where do they submit their nomination? and what are the timelines so people know ahead of time.  (i.e. vacations/life happens)

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+    submitting a brief (1-2 paragraph) application consisting of:
+    *   **Evidence** of performing some of the role's duties over the recent
+        past.
+        *   For SC Member applications, this should include evidence of being
+            well-versed in SLSA.
+    *   **Statement of intent** to continue performing duties over the next
+        year.
+    *   **Affiliation** / employer.
+*   **SC Members decide** if the individual meets the desirable properties
```
melba-lopez, line 84: Shouldn't we allow the community to vote as well for the SC Members or maintainers seat?  SC can have the majority vote, but if you don't have community consensus, this leads to potential bias.

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+    (described in the next section) and respond with an official determination
+    and rationale. Decisions are effective immediately.
+    *   SC Members should attempt to reach consensus. At a minimum:
+        *   SC Member applications require explicit approval from at least
+            50% of the existing SC Members, excluding the candidate.
+        *   Maintainer applications require explicit approval from at least 2
+            existing SC members (possibly including the candidate).
+*   Existing SC Members and Maintainers **annually re-apply** for the position,
```
melba-lopez, line 92: When is the actual timeline?  every year on October... as an example.

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+*   Add SC Member duty: "selecting Maintainers and Steering Committee Members"
+*   Add SC Member duty: "representing SLSA in the OpenSSF SCI WG"
+*   Remove requirement for SC members to meet "no less then [sic] once per
+    quarter".
+*   Add requirement to publish minutes for every SC meeting.
+*   State explicitly that SLSA lives within and takes broad direction from the
+    OpenSSF SCI WG.
+
+## Alternatives considered
```
melba-lopez, line 198: thanks for adding this.

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+    *   These permissions should be reviewed and cleaned up regularly (e.g.
+        annually)
+
+Once this proposal is accepted, the following changes will be made to the
+GitHub organisation:
+
+*   Set up Maintainers teams and add Maintainers as members if needed
+*   Change the org member base permissions to “No permissions” (from “Admin”)
+    and check that everyone still has necessary access
+*   Purge unnecessary permissions in the org and in each repo
```
melba-lopez, line 182: may want to add a cadence.. quarterly.. every 6 months..etc.

0007/README.md
joshuagl: Good questions. I expect we will figure thesse details out as part of the implementation (a PR against [slsa-framework/governance](https://github.com/slsa-framework/governance/)), but my initial suggestion would be:
* nominations are GitHub Issues
* timeline/cadence would be documented in the governance repository and the nomination period would be broadly communicated (in Slack, email, meetings, etc)

0007/README.md
abacchi: For maintiners maybe it should be an open door period rather than one specific timeframe. SC member applicaiton should be a fixed annual application timeline though.

0007/README.md
```diff
@@ -0,0 +1,221 @@
...
+        permission to triage issues or approve PRs
+*   Maintainer: an individual who is **trusted** and **expected to regularly
+    perform** day-to-day duties **for a specific sub-project** within the
+    slsa-framework organization.
+    *   A **Core Maintainer** is a maintainer of the slsa-framework/slsa repo.
+        This repository is special because it contains the core specification
+        and the overall project issue tracker.
+    *   A **Subproject Maintainer** is a maintainer of a different repo within
+        the slsa-framework organization.
+*   SC Member: a **senior leader** of the SLSA project who is **trusted** and
```
abacchi, line 123: It may be nice for the SC group to present/review their activities quarterly / 6 month'ly at the standard SLSA meetings to improve transparency as well.

0007/README.md
joshuagl: Actual timeline tbd. I believe we should test the proposed change to the governance repo before we fully accept it, by running a nomination cycle for the SC. Then will do things annually from that (approximate) date.

Agreed that for maintainers, it should not be a specific timeframe. Good point!

0007/README.md
joshuagl: Good call, I'll add a note that we should do this regularly and at a minimum following each SC nomination cycle.

joshuagl: Added in https://github.com/slsa-framework/slsa-proposals/pull/15/commits/4d91b1ab5415d562e4c3560d3edbbba923070d79

0007/README.md
joshuagl: I like the suggestion and agree with the principal. I am wary of codifying a periodicity, but added some text in [6a96a8c](https://github.com/slsa-framework/slsa-proposals/pull/15/commits/6a96a8cc37d3fe4d33321cacd63d49a58d6aae5e). WDYT?

joshuagl (commented):
Thanks all for the reviews and feedback so far. I've pushed some minor changes in response to review feedback:

* noted that we should perform purging of unneccessary GitHub permissions on a regular cadence
* added SC Member responsibilities:
   * sharing relevant SCI WG happeninings with the community (making this sharing 2-way)
   * that the SC should share updates in the community meeting (I've added a section to the meeting minute template in the Google doc too)

0007/README.md
joshuagl: We have been trying to avoid a formal voting process, as they can be onerous for the people running them. That also helps us avoid the question of how to identify elligible voters.

I hope that the public nomination process will help keep people honest. If we are to use GitHub Issues for nominations, that would allow folks to express support (though admittedly in a non-anonymous way).

MarkLodato: IMO voting is overkill for such a small community. If the community grows 10x in size, then maybe we reconsider?

0007/README.md
MarkLodato: My feeling was rolling, as @abacchi said. We could have some sort of script that sends a notification when a SC member or maintainer's approval is out of date.

0007/README.md
lehors: I think it'd be better to have nominations made in the form of a PR that changes the list of maintainers or SC members as applicable. The PR description can be used to provide whatever information is required and this enforces the maintenance of the list in a publicly visible manner.

0007/README.md
lehors: The biggest challenge in holding such a vote is that it raises the question of who gets to vote. For this reason, I think it is better to keep the decision to the current maintainers and SC members even though it is prone to conflicts of interest.

0007/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Alternatives considered
+
+*   Status quo
+    *   see [goals]
+*   Dissolve the SLSA Steering Committee in favor of steering from the SCI WG
+    *   Con: Strategy may be driven by those without deep familiarity with
+        the project.
+    *   Con: SLSA community may not feel an appropriate sense of ownership and
+        autonomy.
```
lehors: For the sake of completeness this should include the pro:
```suggestion
*   Dissolve the SLSA Steering Committee in favor of steering from the SCI WG
    *   Pro: simplified governance structure
    *   Con: Strategy may be driven by those without deep familiarity with
        the project.
    *   Con: SLSA community may not feel an appropriate sense of ownership and
        autonomy.
```

0007/README.md
```diff
@@ -0,0 +1,225 @@
...
+*   Merge the Core Maintainer and Steering Committee Member roles
+    *   Pro: simplified structure
+        *   Counterpoint: the simplification does not result in too significant
+            a savings
```
lehors: Err... I think you're falling into subjectivity land on this one. :-)

lehors (commented):
Thanks for putting this together. I appreciate the candid approach about the current state we are in and think this is going in the right direction.
I added a few minor comments and have one nit:
The document uses both "OSSF" and "OpenSSF". I know OpenSSF is not consistent on this (we have openssf.org and github.com/ossf) but I think we should stick to OpenSSF which seems to be the dominant form.

brunodom:
@MarkLodato, unfortunately, I can't approve it now because I am traveling
and forgot my security key. I am not able to set up the Authenticator
without logging into my account. Basically, I am lockout these days.
However, consider it approved on my side.

Best Regards!
-Bruno Domingues

0007/README.md
abacchi: yes that update looks great.

0007/README.md
joshuagl: Agreed, added in [4432b4a](https://github.com/slsa-framework/slsa-proposals/pull/15/commits/4432b4a8dd41a8ee2ab73f1a074b409ab7e4e07e). Thanks!

0007/README.md
joshuagl: Fair comment, I've removed this in [4432b4a](https://github.com/slsa-framework/slsa-proposals/pull/15/commits/4432b4a8dd41a8ee2ab73f1a074b409ab7e4e07e).

bobcallaway:
LGTM as well

MarkLodato:
Friendly ping. What do we need to do to get this over the line and accepted?

lehors (approved):

lehors:
> Friendly ping. What do we need to do to get this over the line and accepted?

Who are you pinging and asking this to @MarkLodato ? For what it's worth, I approve. But I think if the current SC approves that should do it since technically the SC owns the governance from what I understand. If you want to be more open about it I suggest you raise it as a proposal to be approved on the next spec group call.

trishankatdatadog:
I would prefer that we get the entire current SC to approve this.

mlieberman85:
I will double check and approve shortly. I've read it, but want to make sure I haven't missed anything.

joshuagl:
It would be good to have approval from the full, current, @slsa-framework/slsa-steering-committee for this proposal. 

- @MarkLodato, @kimsterv and @trishankatdatadog have approved the PR
- @brunodom [approved in a comment]9https://github.com/slsa-framework/slsa-proposals/pull/15#issuecomment-1667159496)
- we're still waiting for @david-a-wheeler and @mlieberman85

mlieberman85 (approved):
LGTM

brunodom (approved):

0007/README.md
david-a-wheeler: You really want an *odd* number when voting, or you risk deadlocks.

0007/README.md
```diff
@@ -0,0 +1,224 @@
...
+    *   Member:
+        *   Used for Maintainers to allow them to be added to a team.
+        *   Used as needed for Participants and Contributors who would like to
+            set up [custom notification routing] and issue assignment across
+            multiple repos.
+        *   Does not grant any special repository permissions, i.e. Base
+            permissions = “No permission”
+    *   Owner:
+        *   Used for SC Members, OpenSSF administrators, and (as needed)
+            Maintainers
```
david-a-wheeler, line 158: I doubt you want that many people as owners. Owners can delete everything.

0007/README.md
david-a-wheeler: I think it's important to have a mechanism for making hard decisions (voting), and then do your best to avoid using it (e.g., by determining why there are differences and working out compromises that address all major concerns).

0007/README.md
david-a-wheeler: We want at least 2 different organizations represented, yes? Shouldn't that be clearly stated?

david-a-wheeler (approved):
It looks okay to me with the following issues:

1. We need to ensure it stays multi-organizational. I believe the current rule is "no more than 2 from 1 organization, just like OpenSSF"; that should continue.
2. Let's ensure there is no deadlock in the voting process. The simple approach is to have an odd number of people; another alternative is to have a tie-breaking rule. But if we have that much contention and *must* make a decision, then we need to have a way to break ties.

joshuagl:
Thanks for the review.

> It looks okay to me with the following issues:
> 
>     1. We need to ensure it stays multi-organizational. I believe the current rule is "no more than 2 from 1 organization, just like OpenSSF"; that should continue.

Completely agree. This requirement is currently part of our documented governance model and shall continue under this proposal.

>     2. Let's ensure there is no deadlock in the voting process. The simple approach is to have an odd number of people; another alternative is to have a tie-breaking rule. But if we have that much contention and _must_ make a decision, then we need to have a way to break ties.

Good call out. The current governance documents require up to 7 participants in the SC. This proposal changes that to "a minimum of 3 and ideally about 5". 
We should explicitly define a tie-breaking, as I think that is more resilient than trying to require an odd number of members of the SC.

PR #16, opened 2023-08-10
Add Proposal 8: Common CI/CD buildType

MarkLodato:
This PR just copies the content from
https://github.com/slsa-framework/slsa/issues/940 so that we have a
place to iterate on the design. The intent is to submit and iterate
quickly rather than having a long-standing pull request.


joshuagl (approved):
I support this proposal and merging the draft and iterating in-tree towards something that can be ACCEPTED.

MarkLodato:
Merging now to let folks iterate. Again, this is not an indication that it is accepted.

